PR #2, opened 2024-12-13
DBZ-8419 Add Debezium Platform helm chart

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8419

replaces: https://github.com/debezium/debezium-platform-conductor/pull/1


jcechace:
@mfvitale Not going to block the merge, just a note that we need to test the examples on MacOS  since you decided to switch to minikube

mfvitale:
> @mfvitale Not going to block the merge, just a note that we need to test the examples on MacOS since you decided to switch to minikube

Well, there is no strict requirement to use minikube, it is just a suggestion since it is just a way to have a *test* cluster.  As far I know both `kind` and `minikube` are fully compatible with all K8s APIs. 

What is important here is to have a k8s cluster with `ingress controller` installed, as highlighted in the *Prerequisites* section of helm README. 

For sure I have tested this with minikube but in terms of helm installation it should be transparent. 

As far I know minikube is compatible also with MacOS. 

Said that, since I have no possibility to test it with a MacOS, it will be good if you can test it with `kind` on MacOS. WDYT?

helm/values.yaml
```diff
@@ -0,0 +1,17 @@
+domain:
+  url: ""
+stage:
+  image: quay.io/debezium/platform-stage:latest
+conductor:
+  image: quay.io/debezium/platform-conductor:nightly
+  offset:
+    existingConfigMap: ""
+kafka:
```
mfvitale, line 9: Wondering if we should completely remove this. @jcechace, @jpechane WDYT?

jcechace: Yea.. What is this supposed to do? I would  deifinitively just create one -- there is no point in having this for helm (unlike the operator for specific DS instances).

jcechace: @mfvitale ok.. that I would remove for sure. 

However, my comment was actually for line 8 and the offset section -- I don't think we need the ability to use an  existing config map in this case.

mfvitale: > However, my comment was actually for line 8 and the offset section -- I don't think we need the ability to use an existing config map in this case.

1. For consistency as we do for the secrets and also for DS operator. 
2. Maybe user want to control the life-cycle and not let the helm remove the offset during an `uninstall`. 

So I would keep it.

jcechace: 1) This has nothing to do with DS operator... completely unrelate.
2) I'm not sure how much sense this would have -- considering it still depends on the WAL retention in the database and all..

However I don't have a strong opinion about it -- we can remove it when it shows to be redundant :)

mfvitale: > I'm not sure how much sense this would have -- considering it still depends on the WAL retention in the database and all..

Currently this is not going to work since the `watcher` is configured without an explicit `snapshot.mode` and the default is `initial`. So if you remove the offset and the re-install the platform the outbox events will be reprocessed. Dunno what really happens on the DS instances, maybe the operator should be able to reconcile without recreating.

mfvitale: > This has nothing to do with DS operator... completely unrelate.

Talking about a more general approach use in the let's call debezium ecosystem. 

> we can remove it when it shows to be redundant :)

And yes, we can always re-work this in future.

PR #8, opened 2025-01-02
DBZ-8512 Support storages supported by Debezium operator for pipeline in Debezium platform

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8512


mfvitale:
Before merging the fix https://github.com/debezium/debezium-operator/pull/81 must be released and the new operator helm chart version must be used.

jcechace:
We should have a call about this.. I have a feeling that once again this is trying to replicate and allow absolutely all possible configurations. The main idea behind the UI is to provided opinionated means. 

Just looking at all the hard coded class and property names looks like an issue in making.

mfvitale:
> I have a feeling that once again this is trying to replicate and allow absolutely all possible configurations.

I think that we talked about to use JDBC as default storage (per deployment) not the only supported one. Isn't it?

We also have another feature in the backlog to let user choose at runtime the preferred storage. 

So with this PR the backend supports all the possible storage supported by the operator and on the other hand the helm chart offers an easy way to provide settings for the JDBC storage. More advanced user can still use another different storage by overriding conductor properties via envs. 

> Just looking at all the hard coded class and property names looks like an issue in making.

This depends on the operator model. Honestly I would have avoided the use of the storage naming and just used the class and the list of generic config just as a pass-trough. This would have simplified the things also in that case. Am I loosing something?

jcechace:
> > I have a feeling that once again this is trying to replicate and allow absolutely all possible configurations.
> 
> I think that we talked about to use JDBC as default storage (per deployment) not the only supported one. Isn't it?
> 
> We also have another feature in the backlog to let user choose at runtime the preferred storage.
> 
> So with this PR the backend supports all the possible storage supported by the operator and on the other hand the helm chart offers an easy way to provide settings for the JDBC storage. More advanced user can still use another different storage by overriding conductor properties via envs.


Sure, but the way to model (and implement it) should be different. It ties back to that "component registry" / "config schema". Storage should be added as a configurable entity -- similar how Transformations are intended to be in the end. Essentially there shouldn't be a hardcoded dependency on particular implementation.

mfvitale:
> Storage should be added as a configurable entity -- similar how Transformations are intended to be in the end. 

This sounds good but then also the DS resource should be able to manage this flexibility. As of now this means to use the custom store from the DS resource in that way the type of storage and the configs can be passed without any logic.  Was you thinking to this approach?

This can be a solution but then why to not have the same simplification on the DS operator?

jcechace:
> > Storage should be added as a configurable entity -- similar how Transformations are intended to be in the end.
> 
> This sounds good but then also the DS resource should be able to manage this flexibility. As of now this means to use the custom store from the DS resource in that way the type of storage and the configs can be passed without any logic. Was you thinking to this approach?
> 
> This can be a solution but then why to not have the same simplification on the DS operator?

I think we should have a call about this since e.g. I don't really follow the note about DS. I will ping you in second half of this week.

mfvitale:
@jcechace can you give another look?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/configuration/TableNameResolver.java
```diff
@@ -0,0 +1,82 @@
...
+
+        String processedValue = PLACE_HOLDERS.entrySet().stream()
+                .reduce(currentValue,
+                        applyReplacementFunction(pipeline),
+                        (lastProcessed, newProcessed) -> newProcessed);
+
+        return sanitizeTableName(processedValue);
+    }
+
+    private BiFunction<String, Map.Entry<String, Function<PipelineFlat, String>>, String> applyReplacementFunction(PipelineFlat pipeline) {
```
jcechace, line 36: What about we don't do that? If you must, create a wrapper class, however this is a bit too much I think :)

jcechace:
@mfvitale sorry for the delay, busy few days. Mostly LGTM except for the one thing I commented on.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/configuration/TableNameResolver.java
mfvitale: What do you think is too much? The function itself?

mfvitale: @jcechace I reworked it. It think it is better readable now but still flexible. WDYT?

jcechace: LGTM... yea, the data type was a bit insane ;)

jcechace (approved):

PR #20, opened 2025-02-03
DBZ-8642: Smart Editor: Add support for the smart editor in the UI for configuring

indraraj:
https://github.com/user-attachments/assets/5fee0e05-8781-4b92-b5ed-f0f6725e79be


indraraj:
FYI @mfvitale

mfvitale:
@indraraj very good enhancement! Is this works also for pipeline and transforms?

Wondering also if we can use this demo video (maybe better arranged with some captions) within some social posts.

indraraj:
> @indraraj very good enhancement! Is this works also for pipeline and transforms?
> 

In pipeline yes, in transform yes in create flow but no in edit flow(still some work needed as properties are not free-text fields but coming from schema) 

> Wondering also if we can use this demo video (maybe better arranged with some captions) within some social posts.

sure, sounds good. are you talking about post captions or in-video text?

PR #24, opened 2025-03-03
DBZ-8640: Fix UI build

indraraj:


mfvitale:
@indraraj Have you checked why it was not caught by the CI?

PR #34, opened 2025-03-24
DBZ-8422: Enable singals from source page UI

indraraj:
- Enable signals in database source setup
- Add a quick filter for notification logs. 

Screenshot 

https://github.com/user-attachments/assets/5c9d4f13-fbd9-4a19-9eaf-4709837f40aa


indraraj:
@mfvitale ^

debezium-platform-stage/src/utils/constants.ts
```diff
@@ -89,6 +89,16 @@ export const transformSchema = {
...
 
+export enum DatabaseType {
+  ORACLE = "ORACLE",
+  MYSQL = "MYSQL",
+  MARIADB = "MARIADB",
+  SQLSERVER = "SQLSERVER",
+  POSTGRESQL = "POSTGRESQL",
+}
+
+export const DEBEZIUM_SIGNAL_COLLECTION_NAME_SUFIX = "_debezium_signal";
```
mfvitale, line 100: IMHO I would avoid to add this suffix and let the user choose the whole name. @Naros @jpechane WDYT?

mfvitale:
@indraraj LGTM, I left only a minor comment.

indraraj:
> @indraraj LGTM, I left only a minor comment.
> 
> Is this PR going to include also the sending of the signal or will be another one?

For sending the signals, I will raise another one.

debezium-platform-stage/src/utils/constants.ts
jpechane: Yes, I agree, we have no such limitation at the core level. I am fine in filling it in input box as starting value but I don't think it should be limited this way.

indraraj: Update to remove the suffix from signal collection name.

PR #35, opened 2025-03-27
DBZ-8851: Add missing oracle and mariadb source

indraraj:
image


mfvitale (approved):

Naros:
@indraraj that looks good to me. 

I'd also suggest in the interim, we replace the MySQL logo with a placeholder. The conditional usage policy doesn't apply to us and thus we need to check internally to see whether we're bound to the same legal consent order as we are with Oracle's logo usage.

indraraj:
Thanks, @Naros updated based on the feedback.

Naros:
@indraraj just a side question, have we thought about our design approach as it relates to i18n? It's certainly not anything to worry with now, but wanted to make sure that if we needed to consider i18n in the future that we have designed the code in a way that can easily be dropped in later.

indraraj:
created this draft [PR](https://github.com/debezium/debezium-platform/pull/36) will close in some time covering all the UI screens.

PR #36, opened 2025-04-02
DBZ-8859: Add the internationalization support in the UI

indraraj:


Naros:
Wow @indraraj I wasn't expecting i18n support so fast :tada: ... Made my week already and it isn't even Monday yet!

PR #37, opened 2025-04-03
DBZ-8857 Remove lombok

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8857


mfvitale:
> @mfvitale LGTM, feel free to merge yourself. One question though, do you think that those classes could be chaned to records?

Unfortunately, Java Records cannot be used as JPA entities since the Records are immutable, and JPA requires the entity class to have a default constructor and be modifiable.

PR #39, opened 2025-04-24
DBZ-8970 Add deletecollection verbs for debeziumserver resource to conductor role

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8970


jpechane:
@mfvitale Applied, thanks

PR #41, opened 2025-04-29
DBZ-8942 Add endpoint to send signals

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8942


jpechane (approved):
@mfvitale Looks good, thanks!
I'd just recommend to use multiline string syntax for sample JSONs as then you'll avoid the need of escaping the double quotes and make it a bit more readable.

Feel free to merge yourself.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/actions/DebeziumServerProxy.java
```diff
@@ -0,0 +1,93 @@
...
+
+        var baseUrl = getDSApiBaseUrl(ds);
+        if (baseUrl.isEmpty()) {
+            throw new DebeziumException("Unable to find pipeline instance to send the signal");
+        }
+
+        var dsApiBaseUrl = baseUrl.get();
+        try (Response response = dsClient.sendSignal(dsApiBaseUrl, signalRequest)) {
+
+            LOGGER.debug("Call to {} returned with {}", baseUrl, response.getStatusInfo().getReasonPhrase());
```
kmos, line 54: I noticed that here, it's logging the `Optional` instead the value. WDYT to pass the value?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/actions/DebeziumServerProxy.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private final KubernetesClient k8s;
+    private final DebeziumServerClient dsClient;
+
+    public DebeziumServerProxy(KubernetesClient k8s, @RestClient DebeziumServerClient dsClient) {
+        this.k8s = k8s;
+        this.dsClient = dsClient;
+    }
+
+    public void sendSignal(SignalRequest signalRequest, DebeziumServer ds) {
```
kmos, line 44: I suggest this kind of proposal:

```java
    public void sendSignal(SignalRequest signalRequest, DebeziumServer debeziumServer) {
        getDSApiBaseUrl(debeziumServer)
                .ifPresentOrElse(url -> send(signalRequest, url),
                        () -> { throw new DebeziumException("Unable to find pipeline instance to send the signal"); } );
    }

    private void send(SignalRequest signalRequest, String dsApiBaseUrl) {
        try (Response response = dsClient.sendSignal(dsApiBaseUrl, signalRequest)) {
            LOGGER.debug("Call to {} returned with {}", dsApiBaseUrl, response.getStatusInfo().getReasonPhrase());
            if (response.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL) {
                LOGGER.error("Sending signal to {} failed with {}", dsApiBaseUrl, response.getStatusInfo().getReasonPhrase());
                throw new DebeziumException(String.format("Unable to to send signal to %s for %s", dsApiBaseUrl, response.getStatusInfo().getReasonPhrase()));
            }
        }
        catch (RuntimeException e) {
            throw new DebeziumException(String.format("Error sending signal to %s ", dsApiBaseUrl), e);
        }
    }
```

the method `send` can be another collaborator.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineController.java
```diff
@@ -236,6 +248,16 @@ public LogReader logReader(Long id) {
         return new KubernetesLogReader(() -> findDeploymentLoggable(id));
     }
 
+    @Override
+    public void sendSignal(Long id, SignalRequest signalRequest) {
+
+        DebeziumServer ds = findById(id)
+                .orElseThrow(() -> new DebeziumException(String.format("Pipeline with id %s not found", id)));
+
+        debeziumServerProxy.sendSignal(signalRequest, ds);
```
kmos, line 257: maybe here we can try to avoid to pass directly the `request` and `map` to a domain object like:

`signalRequest` -> (validation + enrichment) -> `signal`

kmos (approved):

jcechace (changes requested):
Functionality wise LGTM. However this is one of the cases where I would like to see some elementary docs at leas ton class level

- DebeziumServerProxy
- DebeziumServerAttribute

etc ... These are not trivially clear classes and contain literary o documentation.

mfvitale:
@jcechace thanks for your review. Adding the docs forced me to think more about the `KubernetesResourceLocator` and I ended up to rename it to `DebeziumKubernetesAdapter` since now it wraps the `kubernetesClient` and contains the high level operation done on the `DebeziumServer` resource. 

Can you give another look?

jcechace (approved):

PR #49, opened 2025-05-15
DBZ-8422:Sending Signal from UI

indraraj:
Screenshot

https://github.com/user-attachments/assets/67ba8648-d6bf-4eae-80be-b8e44bb7df97


indraraj:
@mfvitale, I am showing these options in the action dropdown. wanted to know if these wordings are fine WDYT?

image

mfvitale:
> @mfvitale, I am showing these options in the action dropdown. wanted to know if these wordings are fine WDYT?
> 
> image

Well I checked on the doc and effectively it is confusing. 

The current action that we have are:

- Log - to print a message in the log with eventually information about the streaming coordinates (offset)
- Ad hoc snapshots with the following action:
   - Start
   - Stop
   - Resume
 The ad hoc snapshots can be of type `INCREMENTAL` or `BLOCKING`.

indraraj:
yes, it's confusing, I tried to get the option's name close to what we have in the doc. 

can you point me to `Resume` action doc in the ad-hoc shapshot is it same as what we have for Incremental snapshot? also what about then `Pause` action is it not available for Ad hoc snapshots

mfvitale:
> can you point me to `Resume` action doc in the ad-hoc shapshot is it same as what we have for Incremental snapshot? 

Yes, sometime we used to use `ad hoc snapshot` and `incremental snapshot` interchangeably but since now we also have the `blocking snapshot` it can lead to confusion. To wrap up, the `ad hoc snapshot` is a a type of snapshot that you can run on demand and the types are: `blocking` or `incremental`.

> what about then `Pause` action is it not available for Ad hoc snapshots
Made a mistake, I updated my previous comment.

indraraj:
@mfvitale updated the options, let me know if any wording need to changed

image

mfvitale:
> @mfvitale updated the options, let me know if any wording need to changed
> 
> image

I'll re-phrase in this way:

* `Add message to the log` -> `Write message to the pipeline log`
* For incremental snapshot
** `Trigger ad hoc incremental snapshot` -> `Start an incremental snapshot`
** `Stop execution of an ad hoc snapshot` -> `Stop an incremental snapshot`
** `Pause ad hoc snapshots` -> `Pause an incremental snapshot`
** `Resume ad hoc snapshots` -> `Resume an incremental snapshot`
* For blocking snapshot
** `Trigger ad hoc blocking snapshot` -> `Start a blocking snapshot`

@Naros WDYT?

PR #50, opened 2025-05-22
DBZ-9045 Enable log notification channel

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9045

### DIFF
--- a/.github/workflows/conductor-workflow.yml
+++ b/.github/workflows/conductor-workflow.yml
@@ -55,6 +55,41 @@ jobs:
           -Dstyle.color=always 
           --no-transfer-progress
           -T4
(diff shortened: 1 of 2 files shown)


mfvitale:
@jpechane is this still blocked for the issue with maven snapshot publishing?

jpechane:
@mfvitale Yes it is. But, should not be operator built also as a part of `Update Dependencies` step?

mfvitale:
> @mfvitale Yes it is. But, should not be operator built also as a part of `Update Dependencies` step?

Well it could.

jpechane:
@mfvitale Applied, thanks

PR #57, opened 2025-07-02
DBZ-8888: Provide support for an auto conversion between formats

indraraj:


indraraj:
@mfvitale FYI

mfvitale:
LGTM! @jpechane @Naros any comments on the UI/UX?

PR #70, opened 2025-08-18
DBZ-9315 Add a rest API to validate a connection

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9315


mfvitale:
> I suppose it will have its own validatior. In that case I'd argue that datase validator should be called database but JDBC or relational database or similar to reflect it is just a subset of databases.

Yes, this could be an option but I leaved it once I'll work on Mongo and eventually refactor.

PR #79, opened 2025-09-09
DBZ-9453 Use maven compiler settings from debezium-parent

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9453


jpechane:
@mfvitale Applied, thanks! I'll trigger the snapshot build to confirm the issue is solved.

PR #353, opened 2026-04-15
debezium/dbz#1816 Hide connection group and move the topic.prefix to diff section

### LINKED ISSUE #1816: Duplicate connection properties in the source connector configuration
Currently, there is duplication of the connector properties in the source connector configuration, as we have a dedicated configuration, plus a configuration group in the connector form.

sol: Hide the configuration group in the form and use the connection entity as configuration.

indraraj:
Closes https://github.com/debezium/dbz/issues/1816

screenshot

https://github.com/user-attachments/assets/07f22372-e601-42c7-8bf7-bcca05472c3e

### DIFF
--- a/debezium-platform-stage/public/locales/en/source.json
+++ b/debezium-platform-stage/public/locales/en/source.json
@@ -32,5 +32,14 @@
         },
         "ddlQuery": "DDL query",
         "errorMsg": "Please select a {{val}} connection in the connector configuration form."
+    },
+    "jumplinks": {
+        "connectorEssentials": "Connector Essentials",
+        "additionalProperties": "Additional Properties"
+    },
+    "form": {
+        "validationFailedGeneric": "Please fill all required fields.",
+        "validationFailedInOneSection": "Please fill all required fields in {{section}} section.",
+        "validationFailedInMultipleSections": "Please fill all required fields in {{list}} sections."
     }
 }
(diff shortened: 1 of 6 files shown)


indraraj:
@mfvitale I have moved the "topic.prefix" field to the Connector Group. WDYT? or should it be moved to the initial block of the properties?

mfvitale:
> @mfvitale I have moved the "topic.prefix" field to the Connector Group. WDYT? or should it be moved to the initial block of the properties?

I think the `Connector` group is right place, we should (in the component descriptors) move the group upper. 

I just noted that maybe we should also hide the `Connection Advanced SSL` since this is related to the `Connection`

indraraj:
> we should (in the component descriptors) move the group upper

By this, you mean changing the groups order in the descriptors api or fixing it in UI?

> I just noted that maybe we should also hide the `Connection Advanced SSL` since this is related to the `Connection`

Sure, also what about the other connection groups like the `Connection Advanced` and `Connection Advanced Replication`? That needs to be hidden as well?

@mfvitale

mfvitale:
> > we should (in the component descriptors) move the group upper
> 
> By this, you mean changing the groups order in the descriptors api or fixing it in UI?

I mean that for now is ok, I'll change it in the descriptors. 

> > I just noted that maybe we should also hide the `Connection Advanced SSL` since this is related to the `Connection`
> 
> Sure, also what about the other connection groups like the `Connection Advanced` and `Connection Advanced Replication`? That needs to be hidden as well?

I'll keep them for now.

indraraj:
updated to hide the `Connection Advanced SSL`

PR #352, opened 2026-04-15
debezium/dbz#1083: Add support for configuring `imagePullPolicy` in H…

pxcamus:
…elm charts for stage and conductor containers, update schema, defaults, and tests accordingly

PR #354, opened 2026-04-16
debezium/dbz#1826 Add support for specify DS image to use

### LINKED ISSUE #1826: Debezium Platform should permit to specify Debezium Server image to be used in the CRD
Currently when a CRD is created based on the pipeline definition, the [image](https://debezium.io/documentation/reference/3.5/operations/debezium-operator.html#debezium-operator-schema-reference-debeziumserverspec-image) is not specified. So the operator will determine the image to use with [ServerImageProvider](https://github.com/debezium/debezium-operator/blob/main/debezium-operator-core/src/main/java/io/debezium/operator/core/ServerImageProvider.java). 

We should add an application property to the `conductor` so the DS image could be specified at deployment time. A default value should be added to `values.yaml` as for the `conductor` and `stage` images.

mfvitale:
closes: debezium/dbz#1826

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/config/PipelineConfigGroup.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/config/PipelineConfigGroup.java
@@ -19,4 +19,6 @@ public interface PipelineConfigGroup {
     @WithName("schema.history")
     SchemaHistoryConfigGroup schema();
 
+    ServerConfigGroup server();
+
 }
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/config/ServerConfigGroup.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/config/ServerConfigGroup.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.platform.config;
+
+import java.util.Optional;
+
+public interface ServerConfigGroup {
+    Optional<String> image();
+}
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
@@ -110,19 +110,23 @@ public class PipelineMapper {
                         this::getPredicateName,
                         this::buildPredicate));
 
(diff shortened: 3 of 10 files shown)


vsantonastaso (commented):
I have only two doubts.
The first one is related to the image validation. What happens if an invalid image is specified? Do we need a sort of validation ?
The second one is related to the security aspects of the custom images. What happens if the custom image is based on private registry with custom authentication or we need secrets or specific policy to pull images?

kmos (approved):

mfvitale:
> What happens if an invalid image is specified? Do we need a sort of validation ? T

It will just breaks. Maybe in future we can thing to add some verification but I never seen this on other helm stuff where you can override images with yours. 

> What happens if the custom image is based on private registry with custom authentication or we need secrets or specific policy to pull images?

This is a good point. 

We recently introduced the `imagePullPolicy` override in https://github.com/debezium/debezium-platform/pull/352 but it was just for `conductor` and `stage` image. We should also extend this to `descriptors` and `debezium-server` image and also introduce the possibility to specify the `imagePullSecrets` too. I prefer to manage this into a separate issue. WDYT?

vsantonastaso:
> > What happens if an invalid image is specified? Do we need a sort of validation ? T
> 
> It will just breaks. Maybe in future we can thing to add some verification but I never seen this on other helm stuff where you can override images with yours.
> 
> > What happens if the custom image is based on private registry with custom authentication or we need secrets or specific policy to pull images?
> 
> This is a good point.
> 
> We recently introduced the `imagePullPolicy` override in #352 but it was just for `conductor` and `stage` image. We should also extend this to `descriptors and `debezium-server`image and also introduce the possibility to specify the`imagePullSecrets` too. I prefer to manage this into a separate issue. WDYT?

Sure

vsantonastaso (approved):

Naros (approved):

mfvitale:
> > > What happens if an invalid image is specified? Do we need a sort of validation ? T
> > 
> > 
> > It will just breaks. Maybe in future we can thing to add some verification but I never seen this on other helm stuff where you can override images with yours.
> > > What happens if the custom image is based on private registry with custom authentication or we need secrets or specific policy to pull images?
> > 
> > 
> > This is a good point.
> > We recently introduced the `imagePullPolicy` override in #352 but it was just for `conductor` and `stage` image. We should also extend this to `descriptors and `debezium-server`image and also introduce the possibility to specify the`imagePullSecrets` too. I prefer to manage this into a separate issue. WDYT?
> 
> Sure

Opened https://github.com/debezium/dbz/issues/1851

PR #355, opened 2026-04-16
debezium/dbz#1825: Add support for enabling `debezium-operator` in He…

### LINKED ISSUE #1825: Add condition to Debezium Platform Chart to conditionally disable operator dependency
Currently the Debezium Platform Helm Chart declares the operator chart has dependency but there is no way to disable it as done for the `database`. 

We should add a `condition: debezium-operator.enabled` to permit to skip the operator installation for those already has it installed.

pxcamus:
…lm charts, update schema, values, chart configuration, and documentation.


pxcamus:
@mfvitale when you have time, thank you!

helm/values.yaml
```diff
@@ -49,4 +49,6 @@ database:
   host: postgres
   auth:
     existingSecret: ""
+debezium-operator:
+  enabled: false
```
mfvitale: @pxcamus Just to maintain the backward compatibility and to add less friction as possible on those wants to install the platform I'd leave it as enabled by default let more advanced users disable if they plan a dedicate installation of the operator. WDYT?

pxcamus: @mfvitale I went with the same default than the db dependency, but it makes sense to have the operator installed by default. Fixed!

mfvitale:
@pxcamus Applied! Thanks a lot!

PR #375, opened 2026-04-27
debezium/DBZ-1768: Add support for configurable labels on DebeziumServer CR

### LINKED ISSUE #1768: Passing labels to debeziumserver Custom resource in Debezium Management Platform

 Is there a way for me to pass/add a label to the debeziumserver Custom resource each time a new pipeline is created? This way I can view all the resources in Argocd UI. I have gone through the CRD in the helm chart, but could not find a way to add a label. Conductor creates the CR, so I am assuming there should be some way for the conductor to add a label to it.

We should be able to view the debeziumserver Custom resource in argocd. To do this need to pass a label to it

---

Binayak490-cyber:
## Summary

Currently, when Conductor creates the DebeziumServer Custom Resource, it only sets one hardcoded label (`debezium.io/conductor-id`). There is no way for users to pass additional labels to the CR, which makes it impossible to group or track resources in tools like ArgoCD.

This PR adds support for configurable labels on the DebeziumServer CR via the pipeline server config.

## Changes

- Added `labels()` method to `ServerConfigGroup` interface
- Updated `PipelineMapper` to extract label building into a `createLabels()` method that merges user-configured labels with the internal `debezium.io/conductor-id` label
- Added `labels: {}` as default empty map in `application.yml`
- Added test coverage for label merging behavior in `PipelineMapperTest`

## Usage

Users can now configure labels via `application.yml` or environment variables:

```yaml
pipeline:
  server:
    labels:
      argocd.argoproj.io/instance: debezium-platform
```
      
## Note

A follow-up issue may be required to be created to track Option C - supporting per-pipeline labels via the API in addition to the config-level labels added here.

Fixes [#1768](https://github.com/debezium/dbz/issues/1768)

### DIFF
--- a/debezium-platform-conductor/src/main/resources/application.yml
+++ b/debezium-platform-conductor/src/main/resources/application.yml
@@ -51,6 +51,7 @@ pipeline:
                 name: "@{pipeline_name}_schema_history"
   server:
     image: ${PIPELINE_SERVER_IMAGE:}
+    labels: {}
 sources:
   database:
     connection:
--- a/debezium-platform-conductor/src/main/resources/application.yml
+++ b/debezium-platform-conductor/src/main/resources/application.yml
@@ -49,9 +49,9 @@ pipeline:
             history:
               table:
                 name: "@{pipeline_name}_schema_history"
+  labels: {}
   server:
     image: ${PIPELINE_SERVER_IMAGE:}
-    labels: {}
 sources:
   database:
     connection:
--- a/helm/templates/conductor-deployment.yaml
+++ b/helm/templates/conductor-deployment.yaml
@@ -73,4 +73,8 @@ spec:
             - name: {{ .name }}
               value: {{ .value }}
 {{- end }}
+{{- range $key, $value := .Values.pipeline.labels }}
+            - name: PIPELINE_LABELS_{{ $key | upper | replace "." "_" | replace "/" "_" | replace "-" "_" }}
+              value: {{ $value | quote }}
+{{- end }}
       restartPolicy: Always
--- a/helm/templates/conductor-deployment.yaml
+++ b/helm/templates/conductor-deployment.yaml
@@ -74,7 +74,7 @@ spec:
               value: {{ .value }}
 {{- end }}
 {{- range $key, $value := .Values.pipeline.labels }}
-            - name: PIPELINE_LABELS_{{ $key | upper | replace "." "_" | replace "/" "_" | replace "-" "_" }}
+            - name: {{ include "debezium-platform.toEnv" (dict "prefix" "PIPELINE_LABELS" "key" $key) | trim }}
               value: {{ $value | quote }}
 {{- end }}
       restartPolicy: Always
--- a/helm/tests/conductor_test.yaml
+++ b/helm/tests/conductor_test.yaml
@@ -258,3 +258,18 @@ tests:
       - equal:
           path: spec.rules[0].host
           value: legacy.example.com
+
+  - it: should configure pipeline labels via env variables
+    template: conductor-deployment.yaml
+    set:
+      env:
+        - name: PIPELINE_LABELS_ARGOCD_ARGOPROJ_IO_INSTANCE
+          value: debezium-platform
+    asserts:
+      - isKind:
+          of: Deployment
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: PIPELINE_LABELS_ARGOCD_ARGOPROJ_IO_INSTANCE
+            value: debezium-platform
--- a/helm/tests/conductor_test.yaml
+++ b/helm/tests/conductor_test.yaml
@@ -262,9 +262,8 @@ tests:
   - it: should configure pipeline labels via env variables
     template: conductor-deployment.yaml
     set:
-      env:
-        - name: PIPELINE_LABELS_ARGOCD_ARGOPROJ_IO_INSTANCE
-          value: debezium-platform
+      pipeline.labels:
+        argocd.argoproj.io/instance: debezium-platform
     asserts:
       - isKind:
           of: Deployment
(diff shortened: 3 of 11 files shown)


Binayak490-cyber:
Hi @jpechane, @mfvitale, @Naros, @kmos, I have raised a PR for this issue implementing **Option A** as discussed. The PR adds support for configurable labels on the `DebeziumServer` CR via `pipeline.server.labels` in the conductor config. These labels are merged with the existing internal `debezium.io/conductor-id` label when the CR is created.

Would appreciate your reviews on this. Thanks!

debezium-platform-conductor/src/main/resources/application.yml
```diff
@@ -51,6 +51,7 @@ pipeline:
                 name: "@{pipeline_name}_schema_history"
   server:
     image: ${PIPELINE_SERVER_IMAGE:}
+    labels: {}
```
mfvitale: I'd like to see this under the `pipeline` since the server here is more a technical thing. WDYT?

mfvitale: Could you then also add a helm unittest?

mfvitale:
@Binayak490-cyber Overall it LGTM. I left a minor comment.

debezium-platform-conductor/src/main/resources/application.yml
Binayak490-cyber: I will go through this and update the PR accordingly.

Binayak490-cyber:
@mfvitale, I've addressed your feedback: moved `labels` to the `pipeline` level in the config and added a Helm unittest to verify it. Let me know if it's good to go!

helm/tests/conductor_test.yaml
```diff
@@ -258,3 +258,18 @@ tests:
       - equal:
           path: spec.rules[0].host
           value: legacy.example.com
+
+  - it: should configure pipeline labels via env variables
+    template: conductor-deployment.yaml
+    set:
+      env:
```
mfvitale: I'm wondering if we could add an explicit `pipeline.labels` directly inside the `values.yaml` so we simply to the user this configuration? Then we use these setting to add the relative ENV to the conductor? WDYT?

Binayak490-cyber: Okk @mfvitale, so then I will expose the `pipeline.labels` directly in `values.yaml` which will provide a much cleaner configuration for users. 

I'll update the Helm template to automatically map `pipeline.labels` to the relative ENV variables for the Conductor, update the unittest, and push the changes shortly.

Binayak490-cyber:
@mfvitale, I've updated the Helm chart to expose `pipeline.labels` directly in `values.yaml`. The Conductor deployment template now automatically maps these labels to the corresponding `PIPELINE_LABELS_*` environment variables. The unittest has been updated accordingly.

Can you please check it out now...!

Users can now configure labels cleanly like this:
```yaml
pipeline:
  labels:
    argocd.argoproj.io/instance: debezium-platform

File: helm/templates/conductor-deployment.yaml [hunk 3]
```diff
@@ -72,5 +72,9 @@ spec:
 {{- range .Values.env }}
             - name: {{ .name }}
               value: {{ .value }}
+{{- end }}
+{{- range $key, $value := .Values.pipeline.labels }}
```
mfvitale, line 76: Maybe we could create an helper function named something like `toEnv` that takes as input the   `key` and a prefix so it can be reused.

mfvitale:
@Binayak490-cyber Great works overall! I just left a minor comment and also we should document the new `pipeline.labels` property in the helm `README`.

helm/templates/conductor-deployment.yaml
Binayak490-cyber: @mfvitale, Done! I've created a reusable `debezium-platform.toEnv` helper in `_helpers.tpl` that takes a `prefix` and `key` as input. The deployment template now uses this helper instead of inline transformation.

Binayak490-cyber:
> @Binayak490-cyber Great works overall! I just left a minor comment and also we should document the new `pipeline.labels` property in the helm `README`.

@mfvitale, I've addressed your feedbacks:

* Created a reusable `debezium-platform.toEnv` helper function in `_helpers.tpl` for env variable name conversion
* Documented `pipeline.labels` in the Helm README (both in the configurations table and a dedicated section with examples)
* Added `pipeline.labels` to `values.schema.json` for schema validation

Can you please take another look at it now...!

mfvitale:
@Binayak490-cyber Let's wait the CI to see if test passes. Just a note, could you please fix the commit message of the first to commit to be compliant?

Binayak490-cyber:
> @Binayak490-cyber Let's wait the CI to see if test passes. Just a note, could you please fix the commit message of the first to commit to be compliant?

@mfvitale, updated the first commit message to follow the expected format (`debezium/dbz#1768`). Please let me know if any further changes are needed.

mfvitale:
@Binayak490-cyber Failure is not related. I'm checking. Meanwhile could you fix also the remaining commit message?

Naros (approved):

Binayak490-cyber:
> @Binayak490-cyber Failure is not related. I'm checking. Meanwhile could you fix also the remaining commit message?

@mfvitale, I’ve updated the remaining commit message to follow the required same format. Please let me know if anything else needs adjustment and can you please check it out now.

mfvitale (approved):

mfvitale:
@Binayak490-cyber One final thing, could you please rebase with main? This should fix the unrelated test failures.

Binayak490-cyber:
> @Binayak490-cyber One final thing, could you please rebase with main? This should fix the unrelated test failures.

@mfvitale, I’ve rebased the branch on the latest main.

mfvitale:
@Binayak490-cyber Applied, thanks!

Binayak490-cyber:
Thanks @mfvitale, @Naros!

PR #381, opened 2026-05-01
debezium/dbz#1871: Add support for mongodb source

### LINKED ISSUE #1871: Adding support for MongoDB source

This feature will enable a user to validate and create a MongoDB connection with the Debezium Platform.

pxcamus:
Fixes debezium/dbz#1871

## PR Checklist


pxcamus:
Hi @mfvitale , WDYT of the approach so far?

mfvitale:
@pxcamus Wondering if we should aim to generalize the `DatabaseInspector` into something like `SourceInspector` since the are other places where we use it for example https://github.com/debezium/debezium-platform/blob/1186605bae7b2d4a2a9c6f9af4ff1f698e8bf127/debezium-platform-conductor/src/main/java/io/debezium/platform/domain/SourceService.java#L56

pxcamus:
> @pxcamus Wondering if we should aim to generalize the `DatabaseInspector` into something like `SourceInspector` since the are other places where we use it for example
> 
> https://github.com/debezium/debezium-platform/blob/1186605bae7b2d4a2a9c6f9af4ff1f698e8bf127/debezium-platform-conductor/src/main/java/io/debezium/platform/domain/SourceService.java#L56

Hi @mfvitale `DatabaseInspector` has two responsibilities right? List JDBC tables, and verify the structure of a table (the only usage I found was for the signalling table). So your idea would be to make the `SourceInspector` abstract ay source. Sounds good :)  would you like to me to try that approach?

pxcamus:
@mfvitale to validate the signalling collection with MongoDB, I am not seeing the mention of a validator document with `$jsonSchema` operator. I don't know of another way to validate that the documents in the collection are properly built. WDYT?

mfvitale:
> $jsonSchema

Well effectively on MongoDB there is not schema so maybe we could just validate by default

pxcamus:
> > $jsonSchema
> 
> Well effectively on MongoDB there is not schema so maybe we could just validate by default

I am referring to this feature: https://www.mongodb.com/docs/manual/core/schema-validation/specify-json-schema/

Invalid documents will be rejected.

mfvitale:
> > > $jsonSchema
> > 
> > 
> > Well effectively on MongoDB there is not schema so maybe we could just validate by default
> 
> I am referring to this feature: https://www.mongodb.com/docs/manual/core/schema-validation/specify-json-schema/
> 
> Invalid documents will be rejected.

@pxcamus It could be an option. The only thing to check if this command can be executed with the [permission](https://debezium.io/documentation/reference/3.5/connectors/mongodb.html#required-user-permissions) that we normally require fro running MongoDB Connector. If so we could think to add this in the validation phase even if then the validation itself will always return `true` right?

The only concern I have is that this deviates from the relational database where we decided to not automatically create the signal table to avoid asking for more database user permission. Even I think that maybe we can change this in future.

pxcamus:
@mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?

pxcamus:
@mfvitale I was thinking of moving `DatabaseCollectionListingProvider.listAvailableCollections(...)` to the `jdbcSourceInspector`, to finish the refactoring. WDYT?

mfvitale:
> @mfvitale I was thinking of moving `DatabaseCollectionListingProvider.listAvailableCollections(...)` to the `jdbcSourceInspector`, to finish the refactoring. WDYT?

Yes, sounds reasonable.

mfvitale:
> @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?

So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?

pxcamus:
> > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> 
> So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?



> > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> 
> So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?

I was thinking: first try and retrieve the `validator` document if it exists. If not, nothing to validate, right? If it does, then perform an insert and validate.

mfvitale:
> > > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> > 
> > 
> > So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?
> 
> > > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> > 
> > 
> > So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?
> 
> I was thinking: first try and retrieve the `validator` document if it exists. If not, nothing to validate, right? If it does, then perform an insert and validate.

The real problem is that the insert to validate will then written on the `oplog` and will then cause the signal to be processed. 

For now I'll go we always valid implementation.

pxcamus:
> > > > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> > > 
> > > 
> > > So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?
> > 
> > 
> > > > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> > > 
> > > 
> > > So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?
> > 
> > 
> > I was thinking: first try and retrieve the `validator` document if it exists. If not, nothing to validate, right? If it does, then perform an insert and validate.
> 
> The real problem is that the insert to validate will then written on the `oplog` and will then cause the signal to be processed.
> 
> For now I'll go we always valid implementation.



> > > > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> > > 
> > > 
> > > So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?
> > 
> > 
> > > > @mfvitale I apologize, I did not communicate clearly enough. I meant to also modify the documentation to encourage users to create the `validator` document, to enforce the JSON schema for the table. if it is there, we can use it, if not we ignore it -- i.e. return `true`?
> > > 
> > > 
> > > So on the `verifySignalDataCollection` yuo just want to create a document to see if it is valid or not?
> > 
> > 
> > I was thinking: first try and retrieve the `validator` document if it exists. If not, nothing to validate, right? If it does, then perform an insert and validate.
> 
> The real problem is that the insert to validate will then written on the `oplog` and will then cause the signal to be processed.
> 
> For now I'll go we always valid implementation.

You mean we always return true for MondoDB signal collection? We can also retrieve the `validator` document (if it exists) and just validate the schema itself locally?

pxcamus:
Hi @mfvitale the last part (before adding tests) is [SignalCollectionVerifyRequest](https://github.com/debezium/debezium-platform/blob/d8cdca2d31b9b62e6c13a5c8642f02cbb2772126/debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/SignalCollectionVerifyRequest.java#L17), which has a private field referencing `DatabaseConnectionConfiguration`. 

I was thinking of replacing that with a `SourceConnectionConfiguration` and a `ConnectionType`, not a `DatabaseType`. It could then be cast specifically to either JDBC connection config, or MongoDB. WDYT?

mfvitale:
> Hi @mfvitale the last part (before adding tests) is [SignalCollectionVerifyRequest](https://github.com/debezium/debezium-platform/blob/d8cdca2d31b9b62e6c13a5c8642f02cbb2772126/debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/SignalCollectionVerifyRequest.java#L17), which has a private field referencing `DatabaseConnectionConfiguration`.
> 
> I was thinking of replacing that with a `SourceConnectionConfiguration` and a `ConnectionType`, not a `DatabaseType`. It could then be cast specifically to either JDBC connection config, or MongoDB. WDYT?



> Hi @mfvitale the last part (before adding tests) is [SignalCollectionVerifyRequest](https://github.com/debezium/debezium-platform/blob/d8cdca2d31b9b62e6c13a5c8642f02cbb2772126/debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/SignalCollectionVerifyRequest.java#L17), which has a private field referencing `DatabaseConnectionConfiguration`.
> 
> I was thinking of replacing that with a `SourceConnectionConfiguration` and a `ConnectionType`, not a `DatabaseType`. It could then be cast specifically to either JDBC connection config, or MongoDB. WDYT?

I'm wondering if we should instead of passing the whole connection config just pass the id, as we do for the `listAvailableCollections`. I think that the `DatabaseConnectionConfiguration` in the `SignalCollectionVerifyRequest` is something that has not been changes after the introduction of the `Connection` entity. 


@indraraj Do you think that it could be easy to send the `connection id` in the signal verify request instead of the whole config? How we actually get the connection info currently?

pxcamus:
> > Hi @mfvitale the last part (before adding tests) is [SignalCollectionVerifyRequest](https://github.com/debezium/debezium-platform/blob/d8cdca2d31b9b62e6c13a5c8642f02cbb2772126/debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/SignalCollectionVerifyRequest.java#L17), which has a private field referencing `DatabaseConnectionConfiguration`.
> > I was thinking of replacing that with a `SourceConnectionConfiguration` and a `ConnectionType`, not a `DatabaseType`. It could then be cast specifically to either JDBC connection config, or MongoDB. WDYT?
> 
> > Hi @mfvitale the last part (before adding tests) is [SignalCollectionVerifyRequest](https://github.com/debezium/debezium-platform/blob/d8cdca2d31b9b62e6c13a5c8642f02cbb2772126/debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/SignalCollectionVerifyRequest.java#L17), which has a private field referencing `DatabaseConnectionConfiguration`.
> > I was thinking of replacing that with a `SourceConnectionConfiguration` and a `ConnectionType`, not a `DatabaseType`. It could then be cast specifically to either JDBC connection config, or MongoDB. WDYT?
> 
> I'm wondering if we should instead of passing the whole connection config just pass the id, as we do for the `listAvailableCollections`. I think that the `DatabaseConnectionConfiguration` in the `SignalCollectionVerifyRequest` is something that has not been changes after the introduction of the `Connection` entity.
> 
> @indraraj Do you think that it could be easy to send the `connection id` in the signal verify request instead of the whole config? How we actually get the connection info currently?

@mfvitale this is what is currently passed when clicking `Validate` in the Stage UI:

For PostgreSQL:
`{"type":"POSTGRESQL","config":{"hostname":"postgresql-rw.databases","port":5432,"username":"debezium","password":"debezium","database":"debezium"},"name":"demo-postgresql-conn"}`

For MongoDB:
`{"type":"MONGODB","config":{"mongodb.connection.string":"mongodb://debezium:debezium@mongodb-1.mongodb-svc.databases.svc.cluster.local:27017/admin?authSource=admin&directConnection=true&serverSelectionTimeoutMS=2000"},"name":"demo-mongodb-conn2"}`

I think the reason is that `Validate` is only available when editing a connection, before saving. `Save changes` is greyed out until the validation is successful. So passing the ID wouldn't work, right?

indraraj:
> SignalCollectionVerifyRequest
> signal verify request
> @indraraj Do you think that it could be easy to send the connection id in the signal verify request instead of the whole config? How we actually get the connection info currently?

@mfvitale Forgive me if I’m not following correctly, but if this is referring to verifying the `signalCollectionName`, currently we are passing the complete connection config object in the verifying call. But we already have this info present with us, as we already make a call to the connection list api to populate the connection dropdown.

But yes, it will be cleaner to just send the `connection id` instead of the whole config.

mfvitale:
> think the reason is that Validate is only available when editing a connection, before saving. Save changes is greyed out until the validation is successful. So passing the ID wouldn't work, right?

@pxcamus I was talking about the `verification` call to check that the signal collection/table is present on the DB. Currently it still passes all the connection details but since this call is done in the `source` creation page after you set the connection, from the list or create one in line, we can just pass the connection id as we do for the `listAvailableCollections`.

pxcamus:
> > think the reason is that Validate is only available when editing a connection, before saving. Save changes is greyed out until the validation is successful. So passing the ID wouldn't work, right?
> 
> @pxcamus I was talking about the `verification` call to check that the signal collection/table is present on the DB. Currently it still passes all the connection details but since this call is done in the `source` creation page after you set the connection, from the list or create one in line, we can just pass the connection id as we do for the `listAvailableCollections`.

@mfvitale got it! I will change that then?

mfvitale:
> > > think the reason is that Validate is only available when editing a connection, before saving. Save changes is greyed out until the validation is successful. So passing the ID wouldn't work, right?
> > 
> > 
> > @pxcamus I was talking about the `verification` call to check that the signal collection/table is present on the DB. Currently it still passes all the connection details but since this call is done in the `source` creation page after you set the connection, from the list or create one in line, we can just pass the connection id as we do for the `listAvailableCollections`.
> 
> @mfvitale got it! I will change that then?

@indraraj If we change the `SignalCollectionVerifyRequest` to have just the id with this PR, do you will be able to easily follow-up with the required changes on the FE?

indraraj:
> @indraraj If we change the `SignalCollectionVerifyRequest` to have just the id with this PR, do you will be able to easily follow-up with the required changes on the FE?

yeah, sure

mfvitale:
@pxcamus You can proceed!

pxcamus:
> > think the reason is that Validate is only available when editing a connection, before saving. Save changes is greyed out until the validation is successful. So passing the ID wouldn't work, right?
> 
> @pxcamus I was talking about the `verification` call to check that the signal collection/table is present on the DB. Currently it still passes all the connection details but since this call is done in the `source` creation page after you set the connection, from the list or create one in line, we can just pass the connection id as we do for the `listAvailableCollections`.

@mfvitale to clarify: if I only pass the connection id, I won't have access to the signalling table name, right? I would need to pass the source id I think.

mfvitale:
> > > think the reason is that Validate is only available when editing a connection, before saving. Save changes is greyed out until the validation is successful. So passing the ID wouldn't work, right?
> > 
> > 
> > @pxcamus I was talking about the `verification` call to check that the signal collection/table is present on the DB. Currently it still passes all the connection details but since this call is done in the `source` creation page after you set the connection, from the list or create one in line, we can just pass the connection id as we do for the `listAvailableCollections`.
> 
> @mfvitale to clarify: if I only pass the connection id, I won't have access to the signalling table name, right? I would need to pass the source id I think.

I mean you just need to use the connection id in replacement of `DatabaseConnectionConfiguration` and the `fullyQualifiedTableName` will remain. Or am I missing something?

pxcamus:
@mfvitale I am going to add tests for MongoDB. Let me know what you think so far, thanks!

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/actions/SignalDataCollectionChecker.java
```diff
@@ -1,64 +0,0 @@
...
-
-import jakarta.enterprise.context.ApplicationScoped;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import io.debezium.platform.environment.database.DatabaseInspector;
-
-@ApplicationScoped
-public class SignalDataCollectionChecker {
```
mfvitale, line 21: Maybe, we could still maintain this and reuse in `JdbcSourceInspector` to better separate the responsibility?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/source/MongoDbConnectionValidator.java
```diff
@@ -0,0 +1,55 @@
...
+@Named("MONGODB")
+public class MongoDbConnectionValidator implements ConnectionValidator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MongoDbConnectionValidator.class);
+
+    @Override
+    public ConnectionValidationResult validate(Connection connectionConfig) {
+        Configuration mongoConfig = toMongoDbConfiguration(connectionConfig);
+
+        String connString = mongoConfig.getString("mongodb.connection.string");
```
mfvitale: Could we use a constant for `mongodb.connection.string`?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/source/MongoDbSourceInspector.java
```diff
@@ -0,0 +1,86 @@
...
+import io.debezium.connector.mongodb.connection.MongoDbConnections;
+import io.debezium.platform.data.dto.CatalogNode;
+import io.debezium.platform.data.dto.CollectionNode;
+import io.debezium.platform.data.dto.CollectionTree;
+import io.debezium.platform.data.dto.SchemaNode;
+import io.debezium.platform.data.dto.SignalDataCollectionVerifyResponse;
+import io.debezium.platform.domain.views.Connection;
+
+@ApplicationScoped
+@Named("MONGODB_SOURCE_INSPECTOR")
```
mfvitale, line 29: Maybe we can just use `MONGODB` and `JDBC` for the other one? It should work since the other `MONGODB`  implements a different interface.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/ConnectionService.java
```diff
@@ -74,42 +65,9 @@ public CollectionTree listAvailableCollections(Long id) {
...
-        try (JdbcConnection jdbcConnection = databaseConnectionFactory.create(databaseConnectionConfiguration)) {
+        SourceInspector sourceInspector = sourceInspectorFactory.getSourceInspector(connectionConfig.getType());
 
-            List<CatalogNode> catalogs = databaseInspector.getAllTableNames(jdbcConnection).entrySet().stream()
-                    .map(this::extractCatalogNode)
-                    .toList();
-
-            return new CollectionTree(catalogs);
-        }
-        catch (Exception e) {
```
mfvitale, line 102: I'll still maintain the exception block here.

mfvitale (commented):
@pxcamus Thanks! Overall It looks good. I left some comments.

debezium-platform-conductor/src/main/resources/connection-schemas.json
```diff
@@ -199,6 +199,46 @@
...
     }
   },
+  {
+    "type": "MONGODB",
+    "schema": {
+      "title": " MongoDb connection properties",
+      "description": " MongoDb connection properties",
+      "type": "object",
+      "required": [
+        "mongodb.connection.string"
```
mfvitale: @pxcamus You should remove the `mongodb` from here since in is already added in https://github.com/debezium/debezium-platform/blob/1be27ac49ac17f8203bcc5b5dde9b1ad447400e9/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java#L221

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/ConnectionService.java
pxcamus: Hi @mfvitale could you please clarify what you mean? Each source has a different context, right? So MongoDB may only have a connection string, no hostname, etc...

Do you want me to move the logic from each implementation to the `catch`? Or implement a custom runtime exception, like `SourceInspectionError`?

mfvitale: Whatever fits well with the refactoring but I would like to catch any potential error during the listing of the collections.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/source/MongoDbSourceInspector.java
pxcamus: I actually got this error:

`CLASS bean [types=[io.debezium.platform.environment.connection.source.MongoDbSourceInspector, io.debezium.platform.environment.connection.source.SourceInspector, java.lang.Object], qualifiers=[@Any, @Default, @Named("MONGODB")], target=io.debezium.platform.environment.connection.source.MongoDbSourceInspector]
CLASS bean [types=[io.debezium.platform.environment.connection.ConnectionValidator, io.debezium.platform.environment.connection.source.MongoDbConnectionValidator, java.lang.Object], qualifiers=[@Any, @Default, @Named("MONGODB")], target=io.debezium.platform.environment.connection.source.MongoDbConnectionValidator]`

debezium-platform-conductor/src/main/resources/connection-schemas.json
pxcamus: @mfvitale I actually had to remove `mongodb` but also add `.`  here:

https://github.com/pxcamus/debezium-platform/blob/a6ddd2189983c7136be4cf385820b555a3594c75/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java#L70

Then it finally worked :)

pxcamus:
@mfvitale ready for another pass, but this time the pipeline is running.

mfvitale (approved):
@pxcamus Great work! LGTM.

pxcamus:
@mfvitale @indraraj I removed my placeholder `collections` from here. But now, this is the shape of the payload in the UI:

`{
    "catalogs": [
        {
            "name": "ecommerce",
            "schemas": [],
            "collections": [
                {
                    "name": "categories",
                    "fullyQualifiedName": "ecommerce.categories"
                },
                {
                    "name": "products",
                    "fullyQualifiedName": "ecommerce.products"
                },
                {
                    "name": "orders",
                    "fullyQualifiedName": "ecommerce.orders"
                },
                {
                    "name": "debezium_signal",
                    "fullyQualifiedName": "ecommerce.debezium_signal"
                }
            ],
            "totalCollections": 4
        }
    ]
}`

Is it possible to skip `schemas` since it is empty for MongoDB?

indraraj:
> @mfvitale @indraraj I removed my placeholder `collections` from here. But now, this is the shape of the payload in the UI:
> 
> `{ "catalogs": [ { "name": "ecommerce", "schemas": [], "collections": [ { "name": "categories", "fullyQualifiedName": "ecommerce.categories" }, { "name": "products", "fullyQualifiedName": "ecommerce.products" }, { "name": "orders", "fullyQualifiedName": "ecommerce.orders" }, { "name": "debezium_signal", "fullyQualifiedName": "ecommerce.debezium_signal" } ], "totalCollections": 4 } ] }`
> 
> Is it possible to skip `schemas` since it is empty for MongoDB?

@pxcamus correct me if I am not getting it right, is this the response for `/api/connections/{id}/collections` api for MongoDB?

I think you can skip it, since we will anyhow be handling the MongoDB data tree separately, as its structure is going to be different from how other relational database responses 

@mfvitale WDYT?

mfvitale:
> > @mfvitale @indraraj I removed my placeholder `collections` from here. But now, this is the shape of the payload in the UI:
> > `{ "catalogs": [ { "name": "ecommerce", "schemas": [], "collections": [ { "name": "categories", "fullyQualifiedName": "ecommerce.categories" }, { "name": "products", "fullyQualifiedName": "ecommerce.products" }, { "name": "orders", "fullyQualifiedName": "ecommerce.orders" }, { "name": "debezium_signal", "fullyQualifiedName": "ecommerce.debezium_signal" } ], "totalCollections": 4 } ] }`
> > Is it possible to skip `schemas` since it is empty for MongoDB?
> 
> @pxcamus correct me if I am not getting it right, is this the response for `/api/connections/{id}/collections` api for MongoDB?
> 
> I think you can skip it, since we will anyhow be handling the MongoDB data tree separately, as its structure is going to be different from how other relational database responses
> 
> @mfvitale WDYT?

I'll let it as is since for example also MySQL/MariaDb do not have the `schema` concept. 

@indraraj What we actually do when `schemas` is empty?

indraraj:
> 
> I'll let it as is since for example also MySQL/MariaDb do not have the `schema` concept.
> 
> @indraraj What we actually do when `schemas` is empty?

For current, we expect all supported source connectors to have a response like,
```
{
  "catalogs": [
    {
      "name": "string",
      "schemas": [
        {
          "name": "string",
          "collections": [
            {
              "name": "string",
              "fullyQualifiedName": "string"
            }
          ],
          "collectionCount": 0
        }
      ],
      "totalCollections": 0
    }
  ]
}
```
 thus an empty schema will imply totalCollections to be  0.

mfvitale:
> > I'll let it as is since for example also MySQL/MariaDb do not have the `schema` concept.
> > @indraraj What we actually do when `schemas` is empty?
> 
> For current, we expect all supported source connectors to have a response like,
> 
> ```
> {
>   "catalogs": [
>     {
>       "name": "string",
>       "schemas": [
>         {
>           "name": "string",
>           "collections": [
>             {
>               "name": "string",
>               "fullyQualifiedName": "string"
>             }
>           ],
>           "collectionCount": 0
>         }
>       ],
>       "totalCollections": 0
>     }
>   ]
> }
> ```
> 
> thus an empty schema will imply totalCollections to be 0.

Go it. So I think that can be situation where the `schemas.name` is `null` and we manage this on the UI right?

indraraj:
> 
> Go it. So I think that can be situation where the `schemas.name` is `null` and we manage this on the UI right?

yes

debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/CatalogNode.java
```diff
@@ -7,9 +7,13 @@
 
 import java.util.List;
 
-public record CatalogNode(String name, List<SchemaNode> schemas, int totalCollections) {
+public record CatalogNode(String name, List<SchemaNode> schemas, List<CollectionNode> collections, int totalCollections) {
```
mfvitale: @pxcamus based on the @indraraj answer, i think you should just set to `null` the name of the schema and just populate the collections, the UI is already capable to manage this situation.

mfvitale (changes requested):
Please take a look to my last comment

debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/CatalogNode.java
indraraj: yes, but to correctly display the collections for MongoDB, work will be needed on the UI side. Since by default it searches the collections inside a schema, which would result in an empty list being shown

mfvitale: > yes, but to correctly display the collections for MongoDB, work will be needed on the UI side. Since by default it searches the collections inside a schema, which would result in an empty list being shown

If the MongDB will return 

```json
{
  "catalogs": [
    {
      "name": "MongoDBDatabase",
      "schemas": [
        {
          "name": null,
          "collections": [
            {
              "name": "MongoDBCollection1",
              "fullyQualifiedName": "MongoDBDatabase.MongoDBCollection1"
            },
              {
              "name": "MongoDBCollection2",
              "fullyQualifiedName": "MongoDBDatabase.MongoDBCollection2."
            }
          ],
          "collectionCount": 2
        }
      ],
      "totalCollections": 2
    }
  ]
}
```

@indraraj Is this going to work? Also because this should be similar to what MySQL/MariaDB should return.

indraraj: Yes, this should work

pxcamus:
@mfvitale  I implemented the change, and everything is working end-to-end, including the UI.

```json
{
    "catalogs": [
        {
            "name": "ecommerce",
            "schemas": [
                {
                    "name": null,
                    "collections": [
                        {
                            "name": "categories",
                            "fullyQualifiedName": "ecommerce.categories"
                        },
                        {
                            "name": "products",
                            "fullyQualifiedName": "ecommerce.products"
                        },
                        {
                            "name": "orders",
                            "fullyQualifiedName": "ecommerce.orders"
                        },
                        {
                            "name": "debezium_signal",
                            "fullyQualifiedName": "ecommerce.debezium_signal"
                        }
                    ],
                    "collectionCount": 4
                }
            ],
            "collections": [],
            "totalCollections": 4
        }
    ]
}
```

image

debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/CatalogNode.java
mfvitale: @pxcamus Is this still required?

pxcamus: No I forgot to remove it, sorry. Done!

mfvitale:
@pxcamus it LGTM! Let's wait for the tests to complete.

One last thing for @indraraj . The singnal verify request now accepts the `connectionId` instead of the full connection properties. If we merge this do you have some cycle to work to the UI part? Just to avoid to have something broken for while. 

In any case I'll wait for the merge after the **3.6.0.Alpha2** will be released on Thu/Fri this week.

indraraj:
> @pxcamus it LGTM! Let's wait for the tests to complete.
> 
> One last thing for @indraraj . The singnal verify request now accepts the `connectionId` instead of the full connection properties. If we merge this do you have some cycle to work to the UI part? Just to avoid to have something broken for while.
> 
> In any case I'll wait for the merge after the **3.6.0.Alpha2** will be released on Thu/Fri this week.

Sure, I will pick this on priority

indraraj:
https://github.com/debezium/debezium-platform/pull/398 this will update the verify payload

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/source/MongoDbSourceInspector.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    private String sanitizeConnectionString(Object connectionString) {
+        if (!(connectionString instanceof String value) || value.isBlank()) {
+            return "<missing MongoDB connection string>";
+        }
+
+        return value.replaceFirst(
+                "^(mongodb(?:\\+srv)?://)([^:/@]+):([^@]+)@",
```
mfvitale: Could you please replace this with a constant?

mfvitale (approved):

mfvitale:
@pxcamus Applied! Thanks a lot!

PR #389, opened 2026-05-08
debezium/dbz#1911 Set the page height to a fixed 100% to fix the css in safari browser

indraraj:
Fixes debezium/dbz#1911

### DIFF
--- a/debezium-platform-stage/src/pages/Destination/Destinations.tsx
+++ b/debezium-platform-stage/src/pages/Destination/Destinations.tsx
@@ -108,7 +108,7 @@ const Destinations: React.FunctionComponent = () => {
   );
 
   return (
-    <div data-tour="destination-page" style={{ flex: 1, display: "flex", flexDirection: "column" }}>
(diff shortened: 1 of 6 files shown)


debezium-platform-stage/src/pages/Destination/Destinations.tsx
```diff
@@ -108,7 +108,7 @@ const Destinations: React.FunctionComponent = () => {
   );
 
   return (
-    <div data-tour="destination-page" style={{ flex: 1, display: "flex", flexDirection: "column" }}>
+    <div data-tour="destination-page" style={{ flex: 1, display: "flex", flexDirection: "column", maxHeight: "100%" }}>
```
mfvitale, line 111: Any reason why we have an in-line style instead of a css class?

indraraj: We follow a mixed approach to styling in our project. Where more complex and reusable stuff are generaly defined using classes, and for simpler ones, are not strictly enforced.

PR #398, opened 2026-05-14
debezium/dbz#1937 Update signal verify payload and use fullyQualifiedName for collections

indraraj:
Fixes debezium/dbz#1937

## Description

- Update signal, verify payload 
- Use fullyQualifiedName for collections 
- Update connection type string for types with no schema


mfvitale:
@indraraj Applied, thanks!

PR #409, opened 2026-05-27
debezium/dbz#1897 Add support for exposing Pipelines metrics to Prometheus via OTel Collector

mfvitale:
Fixes debezium/dbz#1897

## Description
- Add built-in monitoring support to the Debezium Platform using OpenTelemetry and Prometheus
  - When enabled (`monitoring.otel.enabled: true`), the Helm chart deploys an `OpenTelemetryCollector` CR and optionally a `ServiceMonitor` for Prometheus integration
  - The conductor is updated to enable OpenTelemetry metrics export on the DebeziumServer CRDs it creates, allowing pipeline metrics to flow from Debezium Server → OTel
  Collector → Prometheus

  Helm chart changes

  - New `OpenTelemetryCollector` CR template with configurable receivers (OTLP gRPC/HTTP), batch processor, and Prometheus exporter
  - New `ServiceMonitor` template for automatic Prometheus scraping
  - Monitoring env vars (`PIPELINE_MONITORING_OTEL_ENABLED`, `PIPELINE_MONITORING_OTEL_ENDPOINT`) injected into the conductor deployment when monitoring is enabled
  - Helper templates for collector naming and endpoint resolution
  - Updated `NOTES.txt`, `values.schema.json`, and documentation

  Conductor changes

  - New `MonitoringConfigGroup` config interface for OTel settings
  - `PipelineMapper.createRuntime()` conditionally enables OpenTelemetry on the DebeziumServer CRD with a configurable collector endpoint (defaults to `http://localhost:4318`)
  - Unit tests for OTel enabled/disabled/no-endpoint scenarios

 Why the OTel Operator and Prometheus Operator are not included as subcharts

 ### The OTel Operator was initially included as a subchart dependency but removed due to the following reasons:

  - Templated CRD limitation: Helm validates all rendered templates against the Kubernetes API before applying anything (helm#11120
  (https://github.com/helm/helm/issues/11120)). The OTel Operator chart uses templated CRDs (in `templates/` rather than `crds/`), so on helm install the `OpenTelemetryCollector`CRD is not yet registered when Helm tries to validate the collector CR, causing the resource mapping to fail. Unlike the Debezium Operator which places CRDs in `crds/`
  (processed first by Helm), there is no workaround that doesn't involve multiple install steps.
  - Conflicting installations: both operators are cluster-scoped and manage CRDs and admission webhooks across all namespaces. Users who already have these operators in their cluster would face conflicts with duplicate CRDs and webhook registrations.
  - Lifecycle mismatch: operators are typically managed independently with their own upgrade cadence, not tied to the application chart's lifecycle.

  Instead, the chart creates only the CRs (OpenTelemetryCollector, ServiceMonitor) and documents the operators as prerequisites.

## PR Checklist


debezium-platform-conductor/src/main/java/io/debezium/platform/config/MonitoringConfigGroup.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.platform.config;
+
+import java.util.Optional;
+
+public interface MonitoringConfigGroup {
```
Naros, line 20: Let's document the interfaces.

Naros (approved):

vjuranek (commented):
LGTM

vjuranek (approved):

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
```diff
@@ -149,13 +151,28 @@ private Quarkus createQuarkus(PipelineFlat pipeline) {
     }
 
     private Runtime createRuntime() {
+        var metricsBuilder = new MetricsBuilder()
```
kmos: Could you refactor with a strategy this portion of code?

resources/images/pipeline.png
indraraj, line 1: I think we should keep it as it's generally a good idea to show what the UI looks like in the readme.

Naros (approved):

helm/README.md
```diff
@@ -6,6 +6,8 @@ This chart will install the components required to run the Debezium Platform.
 4. [Optional] PostgreSQL database used by conductor to store its data.
 5. [Optional] Strimzi operator: operator for creating Kakfa cluster. In case you want to use a Kafka destination in you
    pipeline.
+6. OpenTelemetry Operator: manages the OpenTelemetry Collector used for pipeline monitoring.
```
pxcamus: **Suggestion (non-blocking):** Just to make it super clear:

```
When monitoring is enabled, the chart also creates:
1. An `OpenTelemetryCollector` custom resource, which requires the OpenTelemetry Operator to already be installed.
2. [Optional] A `ServiceMonitor`, which requires the Prometheus Operator to already be installed.
```

pxcamus (approved):

resources/images/pipeline.png
mfvitale: Good catch. I forget to update the "updated" image with also the "Connections"

helm/README.md
mfvitale: @pxcamus Isn't this covered in the `Monitoring` section below?

pxcamus: Yes, it was just to reinforce, since people (me included) tend to skim the README.

pxcamus (approved):

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/metrics/MetricsExporterStrategyManager.java
```diff
@@ -0,0 +1,45 @@
...
+import io.debezium.operator.api.model.runtime.metrics.MetricsBuilder;
+import io.debezium.platform.config.PipelineConfigGroup;
+
+/**
+ * Manager for applying metrics exporter strategies.
+ * This class discovers all available {@link MetricsExporterStrategy} implementations
+ * via CDI and applies them based on their applicability.
+ */
+@ApplicationScoped
+public class MetricsExporterStrategyManager {
```
kmos: instead of using this class, you could use a jakarta producer that generate the `MetricsExporterStrategy` based on configuration/input parameter. You will have only one bean and you can use directly in `PipelineMapper`.

I think also that you can use an implicit strategy which generate directly the `Runtime`.

mfvitale: Only one point here. At the end, when the full implementation will be completed ( FE part too), there will be only the OTel enabled on the Platform and no more. Do you think improving this still make sense?

In any case I'll investigate the `producer` way.

mfvitale: > instead of using this class, you could use a jakarta producer that generate the MetricsExporterStrategy based on configuration/input parameter. You will have only one bean and you can use directly in PipelineMapper

Currently both strategy are `active` at the same time. Is this something doable with the `producer`?

kmos: > Only one point here. At the end, when the full implementation will be completed ( FE part too), there will be only the OTel enabled on the Platform and no more. Do you think improving this still make sense?
> 
> In any case I'll investigate the `producer` way.

With this in mind, I think it's worth to investigate the creation with a producer of `Runtime` or a class that handle the `Runtime`. In this way when you have to remove the code, you don't have to touch the pipeline but the jakarta producer

mfvitale: I used the `producer` to create the `Metrics` object.

kmos (approved):

PR #422, opened 2026-06-12
debezium/dbz#2027 Enable heartbeat to prevent WAL growing

mfvitale:
Fixes debezium/dbz#2027

## Description
This pull request introduces a heartbeat mechanism to the Debezium platform conductor to ensure the replication slot's confirmed_flush_lsn advances even when there are no outbox events. It adds configuration, database schema, and logic for periodic heartbeat actions, improves logging, and adds integration tests to verify the heartbeat's effect. 

## PR Checklist

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
@@ -17,8 +17,6 @@ import jakarta.annotation.PostConstruct;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.event.Observes;
 
-import org.jboss.logging.Logger;
-
 import io.debezium.config.Configuration;
 import io.debezium.connector.postgresql.PostgresConnector;
 import io.debezium.connector.postgresql.PostgresConnectorConfig;
@@ -26,6 +24,8 @@ import io.debezium.connector.postgresql.PostgresConnectorConfig.AutoCreateMode;
 import io.debezium.embedded.Connect;
 import io.debezium.embedded.EmbeddedEngineConfig;
 import io.debezium.engine.DebeziumEngine;
+import io.debezium.heartbeat.DatabaseHeartbeatImpl;
+import io.debezium.heartbeat.Heartbeat;
 import io.debezium.platform.config.OffsetConfigGroup;
 import io.debezium.platform.environment.watcher.config.WatcherConfig;
 import io.debezium.platform.environment.watcher.consumers.OutboxParentEventConsumer;
@@ -37,10 +37,16 @@ import io.quarkus.runtime.Startup;
 @Startup
 public class ConductorEnvironmentWatcher {
 
+    private static final org.slf4j.Logger LOGGER = org.slf4j.LoggerFactory.getLogger(ConductorEnvironmentWatcher.class);
+
     public static final String CONFIG_PORTION = "\\.config";
     public static final String OFFSET_STORAGE_PREFIX = "offset.storage.";
     public static final String OFFSET_PREFIX = "offset.";
-    private final Logger logger;
+    private static final String HEARTBEAT_DEFAULT_ACTION_QUERY = """
+            INSERT INTO public.heartbeat (id, timestamp) \
+            VALUES (1, now()) \
+            ON CONFLICT (id) DO UPDATE SET timestamp = now()\
+            """;
     private final OutboxParentEventConsumer eventConsumer;
     private final WatcherConfig watcherConfig;
     private final ExecutorService executor = Executors.newSingleThreadExecutor();
@@ -60,7 +66,8 @@ public class ConductorEnvironmentWatcher {
         }
 
         var connection = watcherConfig.connection();
-        var offset = watcherConfig.watcher().offset();
+        var watcher = watcherConfig.watcher();
+        var offset = watcher.offset();
         var outbox = watcherConfig.outbox();
         var extraFields = Stream.of(outbox.aggregateColumn(), outbox.aggregateIdColumn(), outbox.typeColumn())
                 .map(c -> c + ":envelope")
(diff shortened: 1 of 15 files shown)


debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
```diff
@@ -77,23 +86,38 @@ public void start() {
                 .with(PostgresConnectorConfig.DATABASE_NAME, connection.database())
                 .with(PostgresConnectorConfig.PLUGIN_NAME, PostgresConnectorConfig.LogicalDecoder.PGOUTPUT.getValue())
                 .with(PostgresConnectorConfig.INCLUDE_SCHEMA_CHANGES, false)
-                .with(PostgresConnectorConfig.TABLE_INCLUDE_LIST, "public.%s".formatted(outbox.table()))
+                .with(PostgresConnectorConfig.TABLE_INCLUDE_LIST, "public.%s,public.heartbeat".formatted(outbox.table()))
```
Naros, line 90: Do you agree that if the action query can be customized, the table name here should be too? Otherwise we have a few options to avoid disconnected values:

1. Query cannot be customized
2. Query can be customized but is forced to use `public.heartbeat` 
3. Both can be customized allowing for full flexible configuration

wdyt?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
```diff
@@ -77,23 +86,38 @@ public void start() {
...
 
         var config = configurationBuilder.build();
 
-        logger.info("Creating Debezium engine");
+        LOGGER.info("Creating Debezium engine");
         this.engine = DebeziumEngine.create(Connect.class)
                 .using(config.asProperties())
+                .using((success, message, error) -> {
+                    if (error != null) {
+                        LOGGER.error("Debezium engine stopped with error: %s".formatted(message), error);
```
Naros, line 111: Should this use `{}` placeholder rather than `.formatted`?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
```diff
@@ -77,23 +86,38 @@ public void start() {
...
+                .with(DatabaseHeartbeatImpl.HEARTBEAT_ACTION_QUERY_PROPERTY_NAME,
+                        watcher.heartbeat().actionQuery().orElse(HEARTBEAT_DEFAULT_ACTION_QUERY))
                 .with("transforms", "outbox")
                 .with("transforms.outbox.type", EventRouter.class.getName())
-                .with("transforms.outbox.table.fields.additional.placement", extraFields);
+                .with("transforms.outbox.table.fields.additional.placement", extraFields)
+                .with("transforms.outbox.predicate", "isOutboxTable")
+                .with("predicates", "isOutboxTable")
+                .with("predicates.isOutboxTable.type", "org.apache.kafka.connect.transforms.predicates.TopicNameMatches")
+                .with("predicates.isOutboxTable.pattern", ".*\\.%s".formatted(outbox.table()));
```
Naros: This assumes the outbox table name does not contain any special characters. Otherwise, this could have an unintended side effect of matching topics. Should we use `Pattern.quote`?

Naros (commented):
LGTM, just a few inline comments about the configuration setup.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
mfvitale: Yeah, you are right. I'm incline to option 1. @Naros WDYT?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
mfvitale: To pass the `Throwable` the method is `void error(String var1, Throwable var2);` so I think the `.formatted` is right. Isn't it?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/ConductorEnvironmentWatcher.java
mfvitale: @Naros Didi you mean something like 

```java
.with("predicates.isOutboxTable.pattern", Pattern.quote(".*\\.%s".formatted(outbox.table())));
```

debezium-platform-conductor/src/test/java/io/debezium/platform/api/HeartbeatWalAdvancementIT.java
```diff
@@ -0,0 +1,101 @@
...
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import jakarta.inject.Inject;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.org.awaitility.Awaitility;
```
Naros: Accidental shaded Awaitility?

Naros (commented):
Hi @mfvitale LGTM, I found just one other nit-pick. I wonder, should we introduce a rule that fails checkstyle if `org.testcontainers.shaded.*` appear in our imports? This has come up a few times recently. WDYT?

debezium-platform-conductor/src/test/java/io/debezium/platform/api/HeartbeatWalAdvancementIT.java
mfvitale: Fixed. Good catch.

jpechane (approved):

PR #444, opened 2026-06-26
debezium/dbz#2115: Fix connection sequence handling during schema migration

pxcamus:
Add test to validate sequence increment and adjust migration script for version 3.6.0.1.

Fixes debezium/dbz#2115

## Description
Fix connection sequence allocation

The `connection_seq` sequence was created with `increment by 1`, which conflicts with Hibernate's default pooled allocation (size 50) and caused duplicate primary key errors when creating connections.

### Changes
- Added migration `V3.6.0.1__fix_connection_sequence_increment.sql` that aligns `connection_seq` to `increment by 50` and advances it past existing IDs.
- Added a targeted upgrade test in `SchemaEvolutionIT` that migrates up to `3.6.0`, inserts an existing connection, runs the fix, and verifies the sequence increment and next value.

## PR Checklist


pxcamus:
@mfvitale ready for a review when you have time, thanks!

debezium-platform-conductor/src/main/resources/db/migration/V3.6.0.1__fix_connection_sequence_increment.sql
```diff
@@ -0,0 +1,14 @@
+alter sequence connection_seq increment by 50;
```
mfvitale, line 1: We normally use the the same release version, so since the `3.6.0.` is not yet out we should use `3.6.0`. The file already exists so the changes should be appended.

mfvitale (changes requested):

debezium-platform-conductor/src/main/resources/db/migration/V3.6.0.1__fix_connection_sequence_increment.sql
mfvitale: I though a bit and effectively this could be a good solution. So the first 3 number are the release and the fourth is the "oder" of the changes. In this why we can have a descriptive file name. Right?

pxcamus: @mfvitale quick question: if something upgrades, this will not work, correct? I.e. if someone has applied that `src/main/resources/db/migration/V3.6.0__create_heartbeat_table.sql` migration, and now deploys a modified version of it, Flyway will through a checksum error, no? I will try and reproduce that right now.

pxcamus: > I though a bit and effectively this could be a good solution. So the first 3 number are the release and the fourth is the "oder" of the changes. In this why we can have a descriptive file name. Right?

Exactly, and because otherwise you cannot upgrage the platform with an existing database where the same release has been applied.

mfvitale (approved):
LGTM. Let's wait the final release before merging it.

debezium-platform-conductor/src/main/resources/db/migration/V3.6.0.1__fix_connection_sequence_increment.sql
mfvitale: It should be possible to add a `REAMDE` file under the `migration` folder where we can describe the naming convention. Could you verify this?

debezium-platform-conductor/src/main/resources/db/migration/README.md
```diff
@@ -0,0 +1,36 @@
+# Database migrations
+
+This directory contains Flyway SQL migrations for the conductor database.
+
+## Naming convention
+
+Migration files must follow Flyway's versioned migration naming format: 
+```text
+V<version>__<description>.sql
```
mfvitale, line 19: Could we highlight that the `version` is mainly aligned with the debezium current iteration version  and if there are multiple changes in the same version the forth number is just an incremental number for the change?

Could we also mention that we use `versioned migration` and refer to https://www.red-gate.com/blog/flyway-naming-patterns-matter/ somewhere?

pxcamus: Done, I also confirm that those artifacts (`README.md` and `img.png`) are ignored by Flyway pattern detection:

```text
2026-06-29 12:59:16,248 INFO  [org.flywaydb.core.internal.command.DbMigrate] (main) Current version of schema "public": 3.6.0
2026-06-29 12:59:16,254 INFO  [org.flywaydb.core.internal.command.DbMigrate] (main) Migrating schema "public" to version "3.6.0.1 - fix connection sequence increment"
2026-06-29 12:59:16,269 INFO  [org.flywaydb.core.internal.command.DbMigrate] (main) Successfully applied 1 migration to schema "public", now at version v3.6.0.1 (execution time 00:00.010s)
```

mfvitale (approved):
@pxcamus Great work! Thanks. Let's wait the final release and then we can merge.

mfvitale:
@pxcamus Applied, thanks!

PR #446, opened 2026-06-29
debezium/dbz#1427 Duplicate key violations should be more obvious

### LINKED ISSUE #1427: Duplicate key violations should be more obvious [DBZ-9649]
Migrated from [DBZ-9649](https://issues.redhat.com/browse/DBZ-9649)

The user should be better informed when they make a mistake by attempting to create a new record that uses the same named key as an existing record.

For example, right now if I had created a destination named {{Kafka}} and decide to create a second one, rather than the form automatically preventing the submission because the name is used or at the very least telling me that a record exists with the same name on submission, the user is presented with

 !image-2025-11-04-16-08-44-333.png|thumbnail! 

I realize it's a bit more work, but I think in keeping with modern UIs, we likely should handle this in 3 ways:

1. Validate on keystroke and unfocus that the name is permissible and not used.
2. Prevent the create/save buttons from being enabled if the name isn't allowed.
3. On submission, we should return a better error than "Internal Server Error"

KMohnishM:
Fixes debezium/dbz#1427

## Description
Improves name uniqueness validation across all primary Debezium Platform resources (Sources, Destinations, Connections, and Transforms) to prevent generic HTTP 500 "Internal Server Error" messages on duplicate submission and improve overall user experience.

### Backend Changes:
- Introduced a custom `AlreadyExistsException` mapped to HTTP `409 Conflict`.
- Integrated name uniqueness checks in the base `AbstractService.java` for any views implementing `NamedView` before saving or updating.
- Added a safety fallback in `GlobalExceptionMapper.java` to unwrap and catch Hibernate/JDBC unique constraint violations, returning HTTP `409 Conflict` instead of `500 Internal Server Error`.

### Frontend Changes:
- Modified the stage API client functions (`createPost`, `editPut`, and `verifySignals`) to parse detailed error messages (checking for `details` or `error` in response JSON) on failure.
- Enabled secondary client-side validation in `CreateSchemaForm.tsx` (using local fetched lists passed as `existingNames`) to check name uniqueness on keystroke/unfocus, show warning text, and disable submit buttons.
- Applied client-side name uniqueness checks to Connection, Source, Destination, and Transform creation forms.
- Replaced buggy notifications that printed `undefined` for resource names on creation failure by referencing the input payload names directly.

## PR Checklist

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/error/GlobalExceptionMapper.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/error/GlobalExceptionMapper.java
@@ -21,6 +21,10 @@ public class GlobalExceptionMapper {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GlobalExceptionMapper.class);
 
+    private static final java.util.regex.Pattern UNIQUE_CONSTRAINT_PATTERN = java.util.regex.Pattern.compile(
+            "([a-zA-Z0-9]+)_name_key",
+            java.util.regex.Pattern.CASE_INSENSITIVE);
+
     @ServerExceptionMapper(WebApplicationException.class)
     public Response mapWebApplicationException(WebApplicationException ex) {
 
@@ -36,8 +40,7 @@ public class GlobalExceptionMapper {
     public Response mapRuntimeException(RuntimeException ex) {
 
         LOGGER.error("Error while processing request", ex);
-        // Try to unwrap a ConstraintViolationException
-        // Seems that Bean Validation on REST object don't work with Blazebit views.
+        // Try to unwrap unique constraint or validation exceptions
         Throwable cause = ex;
         while (cause != null) {
             if (cause instanceof ConstraintViolationException cve) {
@@ -52,11 +55,36 @@ public class GlobalExceptionMapper {
                 ErrorResponse response = new ErrorResponse("Validation failed", violations);
                 return Response.status(Response.Status.BAD_REQUEST).entity(response).build();
             }
+            if (cause instanceof org.hibernate.exception.ConstraintViolationException hce) {
+                return mapHibernateConstraintViolationException(hce);
+            }
             cause = cause.getCause();
         }
 
         return Response.status(Response.Status.INTERNAL_SERVER_ERROR)
                 .entity(new ErrorResponse("An unexpected error happened", List.of(ex.getMessage())))
+                .build();
+    }
+
+    @ServerExceptionMapper(org.hibernate.exception.ConstraintViolationException.class)
+    public Response mapHibernateConstraintViolationException(org.hibernate.exception.ConstraintViolationException ex) {
+        LOGGER.error("Constraint violation exception occurred", ex);
+
+        String constraintName = ex.getConstraintName();
+        String message = "Resource with this name already exists";
+
+        if (constraintName != null) {
+            java.util.regex.Matcher matcher = UNIQUE_CONSTRAINT_PATTERN.matcher(constraintName);
+
+            if (matcher.find()) {
+                String rawResource = matcher.group(1);
+                String resource = rawResource.substring(0, 1).toUpperCase() + rawResource.substring(1).toLowerCase();
+                message = String.format("%s with this name already exists", resource);
+            }
+        }
+
+        return Response.status(Response.Status.CONFLICT)
+                .entity(new ErrorResponse("Already exists", List.of(message)))
                 .build();
     }
 
--- a/debezium-platform-stage/src/apis/apis.tsx
+++ b/debezium-platform-stage/src/apis/apis.tsx
@@ -233,7 +233,17 @@ export const createPost = async <T,>(
     });
 
     if (!response.ok) {
-      const errorMsg = `Failed to create source: ${response.statusText}`;
+      let errorMsg = `Failed to create source: ${response.statusText}`;
+      try {
+        const errJson = await response.json();
+        if (errJson && errJson.details && errJson.details.length > 0) {
+          errorMsg = errJson.details[0];
+        } else if (errJson && errJson.error) {
+          errorMsg = errJson.error;
+        }
+      } catch {
+        // ignore
+      }
       return { error: errorMsg };
     }
 
@@ -262,7 +272,17 @@ export const editPut = async <T,>(
     });
 
     if (!response.ok) {
-      const errorMsg = `Failed to create source: ${response.statusText}`;
+      let errorMsg = `Failed to create source: ${response.statusText}`;
+      try {
+        const errJson = await response.json();
+        if (errJson && errJson.details && errJson.details.length > 0) {
+          errorMsg = errJson.details[0];
+        } else if (errJson && errJson.error) {
+          errorMsg = errJson.error;
+        }
+      } catch {
+        // ignore
+      }
       return { error: errorMsg };
     }
 
@@ -377,7 +397,17 @@ export const verifySignals = async <T,>(
     });
 
     if (!response.ok) {
-      const errorMsg = `Failed to verify the signals: ${response.statusText}`;
+      let errorMsg = `Failed to verify the signals: ${response.statusText}`;
+      try {
+        const errJson = await response.json();
+        if (errJson && errJson.details && errJson.details.length > 0) {
+          errorMsg = errJson.details[0];
+        } else if (errJson && errJson.error) {
+          errorMsg = errJson.error;
+        }
+      } catch {
+        // ignore
+      }
       return { error: errorMsg };
     }
 
--- a/debezium-platform-stage/src/components/CreateSchemaForm.tsx
+++ b/debezium-platform-stage/src/components/CreateSchemaForm.tsx
@@ -120,6 +120,7 @@ interface CreateSchemaFormProps {
   /** Initial layout; user can still switch via the toggle. Pipeline designer modal uses "tabs". */
   defaultLayoutMode?: "jumplinks" | "tabs";
   hideSignalCollections?: boolean;
+  existingNames?: string[];
 }
 
 export interface CreateSchemaFormHandle {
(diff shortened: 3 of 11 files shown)


indraraj:
Hi @KMohnishM, thanks for your contribution. Can you also add a check for the duplicate pipeline name too

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AbstractService.java
```diff
@@ -97,6 +99,9 @@ public <V> Optional<V> findByIdAs(Class<V> viewType, Long id) {
     }
 
     public T create(@Valid T view) {
+        if (view instanceof NamedView namedView) {
+            checkNameUniqueness(namedView.getName(), null);
```
mfvitale: Are we are to directly check the duplicate key exception from `PersistenceException`? In this way can just relay on database uniqueness check?

KMohnishM: Hi @mfvitale, 

Yes, we can directly catch the unique constraint violation exception from the database (and we actually did implement a safety fallback/unwrapper in `GlobalExceptionMapper.java` to prevent database exceptions from leaking as 500 errors).

However, relying **solely** on the database constraint violation exception presents a few challenges:

1. **Brittle Exception Parsing**: Uniqueness constraint violations and their error messages are highly database-dialect-specific (e.g., PostgreSQL vs. H2 vs. Oracle). Trying to parse the raw database exception to extract a clean entity name and value in a database-independent way is complex and brittle.
2. **UX / Clean API Responses**: Performing the pre-insertion check in `AbstractService.java` allows us to raise a clean, localized `AlreadyExistsException` (e.g., `"Source with name '...' already exists"`) before attempting the operation. This keeps the JAX-RS API responses consistent and user-friendly.
3. **Defense in Depth**: The hybrid approach (pre-insertion check + mapper fallback) gives us the best of both worlds:
   - The pre-check SELECT serves as the primary validator to provide consistent, clean validation messages.
   - The `GlobalExceptionMapper` fallback serves as a secondary guard to handle concurrent insert race conditions gracefully (returning HTTP 409 Conflict instead of an unhandled HTTP 500).

Let me know if you would still prefer removing the pre-check query, but this hybrid setup is generally safer for dialect-agnostic clean error messaging!

mfvitale: > 1. **Brittle Exception Parsing**: Uniqueness constraint violations and their error messages are highly database-dialect-specific (e.g., PostgreSQL vs. H2 vs. Oracle). Trying to parse the raw database exception to extract a clean entity name and value in a database-independent way is complex and brittle.

The only supported database for now is `PostgreSQL`



> 2\. **UX / Clean API Responses**: Performing the pre-insertion check in `AbstractService.java` allows us to raise a clean, localized `AlreadyExistsException` (e.g., `"Source with name '...' already exists"`) before attempting the operation. This keeps the JAX-RS API responses consistent and user-friendly.

Can't you just catch the right exception and the provide the useful message to the UI?

KMohnishM: Makes sense @mfvitale, I have removed the pre-check query (`checkNameUniqueness`) from `AbstractService.java` and replaced it entirely with a database-level unique constraint handler in `GlobalExceptionMapper.java`.

The exception mapper now:
1. Intercepts `org.hibernate.exception.ConstraintViolationException`.
2. Inspects the violated constraint name to map it back to the clean resource name (Connection, Source, Destination, Pipeline, Transform, Vault).
3. Safely extracts the duplicate key name from the PostgreSQL server error detail via reflection (`getServerErrorMessage().getDetail()`), preventing compile-time dependencies on the JDBC driver.
4. Returns the clean user-facing conflict message (e.g., `"<Resource> with name '<name>' already exists"`) mapped to HTTP 409 Conflict.

debezium-platform-stage/src/components/pipelineDesigner/PipelineDestinationModel.tsx
```diff
@@ -50,7 +50,12 @@ const PipelineDestinationModel: React.FC<PipelineDestinationModelProps> = ({
     const response = await fetchData<CatalogApiResponse>(
       `${API_URL}/api/catalog`
     );
-    return (response.components["server-sink"] ?? []).map((entry) => ({
+    const list = [
+      ...(response.components["server-sink"] ?? []),
+      ...(response.components["sink-connector"] ?? []),
+      ...(response.components["unknown"] ?? []),
```
indraraj: https://github.com/debezium/dbz/issues/2192 This might be an unintended side effect of some recent server-side work. We should avoid introducing a temporary workaround in the UI

indraraj (changes requested):

debezium-platform-stage/src/components/pipelineDesigner/PipelineDestinationModel.tsx
KMohnishM: Makes total sense! I have completely reverted the temporary UI catalog workaround.

The PR has been force-pushed with a clean, signed-off commit containing only the duplicate key violation fixes for Debezium Platform. We will wait for the server-side fix for https://github.com/debezium/dbz/issues/2192 to resolve the E2E checks.

Thank you!

indraraj: Looks good now. Can you also add these checks to the edit flow. Plus, use i18n for any static text

KMohnishM: Hi @indraraj,

Done! I have added the name uniqueness checks to all the edit flows (Sources, Destinations, Connections, and Transforms) and wrapped all the new static validation messages in t(...) translation functions for i18n.

I've force-pushed the updated commit. Please let me know if everything looks good now!

indraraj (approved):
Looks good, thanks

debezium-platform-conductor/src/main/java/io/debezium/platform/error/AlreadyExistsException.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.platform.error;
+
+public class AlreadyExistsException extends RuntimeException {
```
mfvitale: Is this still used?

debezium-platform-conductor/src/main/java/io/debezium/platform/error/GlobalExceptionMapper.java
```diff
@@ -55,11 +65,95 @@ public Response mapRuntimeException(RuntimeException ex) {
...
+    public Response mapHibernateConstraintViolationException(org.hibernate.exception.ConstraintViolationException ex) {
+        LOGGER.error("Constraint violation exception occurred", ex);
+
+        String constraintName = ex.getConstraintName();
+        String message = "Resource with this name already exists";
+
+        if (constraintName != null) {
+            String normalized = constraintName.toLowerCase();
+            String resource = null;
+            if (normalized.contains("connection")) {
```
mfvitale: Is it possible to extract the resource name with a regex?

debezium-platform-conductor/src/main/java/io/debezium/platform/error/GlobalExceptionMapper.java
```diff
@@ -55,11 +65,95 @@ public Response mapRuntimeException(RuntimeException ex) {
...
+            }
+            else if (normalized.contains("vault")) {
+                resource = "Vault";
+            }
+
+            if (resource != null) {
+                String detail = null;
+                Throwable cause = ex.getCause();
+                while (cause != null) {
+                    if ("org.postgresql.util.PSQLException".equals(cause.getClass().getName())) {
```
mfvitale: Do we really need all the details? I think it is enough saying that there is a duplicate?

KMohnishM:
Thanks for the feedback @mfvitale ! I have updated the PR according to the feeback , Please let me know what you think when you have a moment.

debezium-platform-conductor/src/main/java/io/debezium/platform/error/GlobalExceptionMapper.java
```diff
@@ -55,11 +55,44 @@ public Response mapRuntimeException(RuntimeException ex) {
             cause = cause.getCause();
         }
 
+        // Safety fallback: unwrap to check for Hibernate unique constraint violations
+        cause = ex;
+        while (cause != null) {
+            if (cause instanceof org.hibernate.exception.ConstraintViolationException hce) {
```
mfvitale, line 58: Why not adding this if to the loop above?

debezium-platform-conductor/src/main/java/io/debezium/platform/error/GlobalExceptionMapper.java
```diff
@@ -55,11 +55,44 @@ public Response mapRuntimeException(RuntimeException ex) {
...
 
+    @ServerExceptionMapper(org.hibernate.exception.ConstraintViolationException.class)
+    public Response mapHibernateConstraintViolationException(org.hibernate.exception.ConstraintViolationException ex) {
+        LOGGER.error("Constraint violation exception occurred", ex);
+
+        String constraintName = ex.getConstraintName();
+        String message = "Resource with this name already exists";
+
+        if (constraintName != null) {
+            java.util.regex.Matcher matcher = java.util.regex.Pattern.compile(
```
mfvitale: maybe you can move the `Pattern` as  static field calling it with a proper name?

mfvitale:
@KMohnishM Overall LGTM. I left some minor comment to improve a bit the code.

mfvitale:
@indraraj Could you give a final look.

mfvitale (approved):

mfvitale:
If CI passes it can be merged from my point of view.

KMohnishM:
@mfvitale @indraraj All CI checks have passed now. If everything looks good from your side, I believe this PR is ready to be merged. Thanks !

indraraj:
looks good, we can merge it 👍

PR #453, opened 2026-07-06
Dbz 2125 feedback frontend pipeline fail

pxcamus:
Fixes debezium/dbz#2125

## Description
Adds pipeline deployment failure feedback for skipped non-retriable outbox events.

This change introduces pipeline status and errorMessage fields, persists them in the database, exposes them through the pipeline REST API, and updates failed pipeline deployments to FAILED when outbox event processing is skipped after a non-retriable error. This allows the frontend to display deployment failures and their error details to users.

Also adds tests covering failure-state persistence/API exposure and the outbox deployment failure path.

## PR Checklist


debezium-platform-conductor/src/test/java/io/debezium/platform/MockedTestProfile.java
```diff
@@ -13,4 +15,15 @@ public class MockedTestProfile implements QuarkusTestProfile {
     public String getConfigProfile() {
         return "test"; // optional, for config resolution
     }
+
```
pxcamus: @mfvitale I spent a lot of time on this, but could not get past this error when running the tests:

```bash
Caused by: com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: {"message":"failed to set up container networking: driver failed programming external connectivity on endpoint affectionate_moser (004dcdda3dad16a70e5265272de8997be46f93317922039a6c356e77e1c686b6): Bind for 0.0.0.0:25000 failed: port is already allocated"}
```

This is profile does not use oras at all, I thought it would be ok to disable it?

debezium-platform-conductor/src/test/java/io/debezium/platform/api/PipelineResourceIT.java
```diff
@@ -289,6 +295,8 @@ void createPipelineWithoutServerImage() {
                 .pollInterval(Duration.of(500, ChronoUnit.MILLIS))
                 .untilAsserted(() -> {
                     Mockito.verify(k8sAdapter, Mockito.atLeastOnce()).deployPipeline(debeziumServerArgumentCaptor.capture());
+                    assertThat(debeziumServerArgumentCaptor.getAllValues().stream()
```
pxcamus, line 302: @mfvitale Here the wait was passing on some other tests, because I was consistently getting a `NoSuchElementException` in this path:
```Java
.findFirst()
.orElseThrow();
```

So I tried making it wait for the exact pipeline name.

pxcamus:
@mfvitale Ready when you have time, thanks!

debezium-platform-conductor/src/main/resources/db/migration/V3.7.0__add_pipeline_status.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE pipeline
```
mfvitale, line 1: This should be `V3.7.0` now

debezium-platform-conductor/src/test/java/io/debezium/platform/MockedTestProfile.java
```diff
@@ -13,4 +15,15 @@ public class MockedTestProfile implements QuarkusTestProfile {
...
+    @Override
+    public Map<String, String> getConfigOverrides() {
```
mfvitale: Could you try to remove this? Let's see the CI result then.

debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/PipelineStatus.java
```diff
@@ -6,14 +6,14 @@
...
- * Represents the runtime status of a pipeline deployment on a remote host.
+ * Represents the runtime status of a pipeline deployment.
  *
  * <ul>
- *   <li>{@code DEPLOYING} — Docker container creation requested, awaiting confirmation</li>
- *   <li>{@code RUNNING} — container is running normally</li>
- *   <li>{@code STOPPED} — container was explicitly stopped by the user</li>
- *   <li>{@code FAILED} — container exited unexpectedly or Agent is unreachable</li>
- *   <li>{@code CONFIG_DRIFT} — SHA-256 hash of deployed config does not match expected hash</li>
+ *   <li>{@code DEPLOYING} — deployment creation or update has been requested and is awaiting confirmation</li>
```
mfvitale: I'll use "pipeline" as a more high level concept here. 

Maybe we could also update the `DeploymentStatus` to be `PipelineStatus`

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/PipelineService.java
```diff
@@ -93,4 +94,14 @@ public Optional<String> send(Long pipelineId, Signal signal) {
                     return signal.id();
                 });
     }
+
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public void markFailed(Long pipelineId, String errorMessage) {
```
mfvitale: Maybe we could let this be more generic passing the `status` as parameter? The method could be called `updateStatus`?

debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/PipelineStatus.java
pxcamus: @mfvitale I’ll make the pipeline-facing status `PipelineStatus`. Quick clarification: should lower-level host deployment state keep the deployment terminology, since it represents an actual runtime deployment on a host, or should we align that naming with `PipelineStatus` too?

If yes, I will create a separate enum for `PipelineStatus`.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/watcher/consumers/OutboxParentEventConsumer.java
```diff
@@ -85,12 +89,19 @@ private void consumeWithRetry(EnvironmentEventConsumer<?> consumer, EventContext
                             context.eventType(), context.aggregateType(), context.aggregateId(),
                             attempt > 1 ? " after %d retries".formatted(attempt - 1) : "",
                             e);
+                    markPipelineFailed(context, e);
```
mfvitale: Maybe we could do a little bit better here to avoid adding more responsability to this class that is just a router. 

What if we push the failure-handling responsibility into the `EnvironmentEventConsumer` interface itself?

You could add an `onError(Long id, String aggregateType, String eventType, Exception e)` default method to `EnvironmentEventConsumer`and let `PipelineConsumer` override it to call `pipelineService.markFailed()`. 
The parent consumer would then just delegate the error to the matching consumer(s) instead of handling it directly.

debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/PipelineStatus.java
mfvitale: This is a good point. Effectively the `CONFIG_DRIFT` makes sens only for host based deployment. For now I will keep it and just rename the thinks to be `PipelineStatus`. Then I'll see if we could avoid to persist the `CONFIG_DRIFT` and just act upon the detection. 

@d1vyanshu-kumar Do you think we strictly need to have the `CONFIG_DRIFT` status saved?

debezium-platform-conductor/src/test/java/io/debezium/platform/MockedTestProfile.java
pxcamus: It worked!

pxcamus:
@mfvitale ready for another round  :)

debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/PipelineStatus.java
d1vyanshu-kumar: @mfvitale @pxcamus, Good question! Since I've been working on the host-based deployment design ([DDD-41](https://github.com/debezium/debezium-design-documents/pull/45)), here's my take:

No, I don't think we need to persist `CONFIG_DRIFT `as a pipeline-level status. It's a deployment-infrastructure concern, not a user-facing pipeline state.

In the host-based deployment design, `CONFIG_DRIFT` is detected by a status poller that compares SHA-256 hashes of the deployed application.properties against what the Conductor expects. When drift is detected, the correct response is to act on it -- re-deploy with the correct config automatically, rather than surface it as a persistent status to the user.

At the pipeline level, the user cares about: Is my pipeline deploying? Running? Failed? They don't need to know about internal config hash mismatches.

I'd suggest keeping `CONFIG_DRIFT` only at the host deployment level (in HostDeploymentEntity) where it belongs, and keeping the pipeline-level PipelineStatus clean: DEPLOYING, RUNNING, STOPPED, FAILED.

That said, I have one concern about this PR: merging the old DeploymentStatus (which I introduced for HostDeploymentEntity in the host-deployment work) into PipelineStatus. In my current branch, HostDeploymentEntity uses DeploymentStatus for the low-level container state on the remote host. If we could keep a separate enum for the host deployment (even if it shares some values), it would avoid coupling the pipeline-level status with infrastructure-level details. But I can adapt to whatever approach you prefer -- just flagging the overlap so we don't run into merge conflicts later.

mfvitale: Well, with things starts to be complicated :D

Since the idea for the host-based is to be an alternative to the k8s, and so this is way we track the `Deployment` while for k8s this is managed by the cluster. Probably in future, we could have the concept of deployment also for K8s.

For now, given where we are now. Maybe the right solution is to leave the  `DeploymentStatus` as is and only used for the host-based, for the k8s we can think to add the `RUNNING, STOPPED, FAILED`, where with this PR we just set the `FAILED` status.

@pxcamus @d1vyanshu-kumar WDYT?

pxcamus: That sounds like the cleanest compromise. I’ll keep `DeploymentStatus` for the host-based deployment model, where `CONFIG_DRIFT` belongs, and introduce/use `PipelineStatus` only for the pipeline-facing status exposed by this PR, if you guys are ok with that.

mfvitale:
@pxcamus WDYT If we add a `PipelineLifecycleListener` like 

```java
public interface PipelineLifecycleListener {

    void onRunning(Long pipelineId);

    void onStopped(Long pipelineId);

    void onFailed(Long pipelineId, String errorMessage);
}
```

Then a default implemementation that will use the current `updateStatus` code while in the `PipelineService` the  `updateStatus` will call the relative method on the listener for all available listeners implementations. 

WDYT?

pxcamus:
> @pxcamus WDYT If we add a `PipelineLifecycleListener` like
> 
> ```java
> public interface PipelineLifecycleListener {
> 
>     void onRunning(Long pipelineId);
> 
>     void onStopped(Long pipelineId);
> 
>     void onFailed(Long pipelineId, String errorMessage);
> }
> ```
> 
> Then a default implemementation that will use the current `updateStatus` code while in the `PipelineService` the `updateStatus` will call the relative method on the listener for all available listeners implementations.
> 
> WDYT?

@mfvitale I do have a question: should we not let persistence go through first and then fire the status change once the enitty is updated? Otherwise what would happen if persistence is one listener amongst others, and the transaction has to rollback? 

So the listener is absolutely the right pattern, but I would recommend making the write happen first and the notification fire after the commit. WDYT?

mfvitale:
> > @pxcamus WDYT If we add a `PipelineLifecycleListener` like
> > ```java
> > public interface PipelineLifecycleListener {
> > 
> >     void onRunning(Long pipelineId);
> > 
> >     void onStopped(Long pipelineId);
> > 
> >     void onFailed(Long pipelineId, String errorMessage);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then a default implemementation that will use the current `updateStatus` code while in the `PipelineService` the `updateStatus` will call the relative method on the listener for all available listeners implementations.
> > WDYT?
> 
> @mfvitale I do have a question: should we not let persistence go through first and then fire the status change once the enitty is updated? Otherwise what would happen if persistence is one listener amongst others, and the transaction has to rollback?
> 
> So the listener is absolutely the right pattern, but I would recommend making the write happen first and the notification fire after the commit. WDYT?

Yes, I agree.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/PipelineService.java
```diff
@@ -93,4 +97,15 @@ public Optional<String> send(Long pipelineId, Signal signal) {
                     return signal.id();
                 });
     }
+
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public void updateStatus(Long pipelineId, PipelineStatus status, String errorMessage) {
```
mfvitale, line 102: What about the other non failing status?

mfvitale:
@pxcamus Could you please rebase with main?

pxcamus:
> @pxcamus Could you please rebase with main?

Done @mfvitale

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/PipelineService.java
pxcamus: @mfvitale I left the others out on purpose. There's no feedback loop for the non-failing states yet, right? 

`OperatorPipelineController.deploy()` just applies the DebeziumServer CR and returns (fire-and-forget), and nothing reads the server/pod state back. So the only transition for which I have an actual transition is `failed` on the exception path.
  
If I set `DEPLOYING` at deploy time, nothing would ever move it to `RUNNING`, right? 
  
Is that in scope for this PR, or a follow-up? And are there existing signals that I am missing?

mfvitale: Yeah, we can leave the other states in a separate issue. I think it could fall inside the "Dispatch K8 events from Debezium operator" task we already have in the roadmap.

mfvitale:
@pxcamus Applied, thanks!

mfvitale:
@indraraj Could you create an issue for the `stage` to use the pipeline status and show an error on the pipeline? I'd also add a retry button to just send an update call.

indraraj:
https://github.com/debezium/dbz/issues/2261

PR #460, opened 2026-07-09
debezium/dbz#2090 Add SSH config file watcher with host reconciliation

d1vyanshu-kumar:
Fixes debezium/dbz#2090

## Description

Adds the SSH config file watcher that monitors `~/.ssh/config` and keeps the `host_status` database table in sync automatically.

**How it works:**

`SshConfigWatcherService` registers a Java NIO `WatchService` on the SSH config parent directory and filters events to the `config` file specifically. A 2-second debounce coalesces rapid editor saves into a single reconciliation run. A `@Scheduled(every = "5m", delayed = "5m")` fallback handles environments where `WatchService` misses events — such as NFS mounts and Kubernetes ConfigMap volumes, which use atomic symlink-swap instead of `ENTRY_MODIFY` events.

**Reconciliation logic:**

`HostReconciliationPlanBuilder` is a dedicated, package-private class that compares SSH config state against database state and produces a `ReconciliationPlan` — three lists describing what to add, remove, or update. It is intentionally not a CDI bean: no injected dependencies, no mutable state, no side effects. This makes it directly
unit-testable without a running Quarkus application.

**Persistence layer:**

`HostStatusService` extends `AbstractService<HostStatusEntity, HostStatus, HostStatusReference>` following the same Blaze-Persistence pattern used by `PipelineService`, `VaultService`, and `ConnectionService`. The `HostStatus` entity view is annotated `@CreatableEntityView` and `@UpdatableEntityView` to support all three operations the reconciler needs.

**CDI transaction boundary:**

`@Transactional` on self-invocations is silently ignored. All database write methods live in the separate `HostStatusService` bean so every call goes through the CDI proxy and transactions are properly committed and rolled back.

**Runtime mode guard:**

The bean is always instantiated. Each `@Observes` and `@Scheduled` method is guarded by `if (!isHostMode()) return;` so no file system access happens in operator mode.

**`HostProvisioningService` stub:**

The watcher calls `provisioningService.provision(alias)` for new and updated hosts. The class is included here with empty method bodies so this sub-issue compiles independently. Ansible execution will be implemented in a future implementation.

**Tests:**

`HostReconciliationPlanBuilderTest` contains 7 plain JUnit 5 tests — no `@QuarkusTest`, no database, no mocks of the watcher itself. It directly instantiates the builder and verifies every reconciliation scenario: add, remove, update, mixed, empty, unchanged,
and null-hostname fallback.

## PR Checklist

- [X] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md)
  and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [X] Minimal changes to code not directly related to your change
- [X] One feature/change per PR
- [X] Rebased on upstream `main`


debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/HostReconciliationRepository.java
```diff
@@ -0,0 +1,110 @@
...
+ * {@code @Transactional} annotations are ignored on self-invocations within
+ * the same CDI bean. The {@link SshConfigWatcherService} delegates all DB
+ * writes to this repository so that each call goes through the CDI proxy
+ * and the transaction is properly managed.
+ *
+ * <p>Uses {@link EntityManager} directly, matching the existing project pattern
+ * (see {@code AbstractService}, {@code PipelineService}, etc.).
+ */
+@ApplicationScoped
+public class HostReconciliationRepository {
```
mfvitale: Though I like the repository pattern, for historic reason, we use the `BlazeBit` persistence layer. 
Could you please so align with it? This means your service will extend the `AbstractService`, so same basic method like `list, create, update, delete, findBy` are already there.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
```diff
@@ -0,0 +1,438 @@
...
+     *
+     * <p>This method is a <strong>pure function</strong> — it has zero database
+     * access and zero side effects. It exists as package-visible so that unit
+     * tests can exercise the reconciliation logic without a running database.
+     *
+     * @param fileHosts hosts parsed from the SSH config file
+     * @param dbHosts   active host entities from the database (status ≠ REMOVED)
+     * @return the reconciliation plan describing what needs to change
+     */
+    ReconciliationPlan buildReconciliationPlan(List<SshHostEntry> fileHosts,
```
mfvitale: WDYT to put this on a separate class?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
```diff
@@ -0,0 +1,438 @@
...
+
+    /**
+     * Schedules a reconciliation after the debounce delay. Cancels any pending
+     * reconciliation so that multiple rapid file events result in exactly
+     * one reconciliation run.
+     */
+    private void scheduleReconciliation() {
+        ScheduledFuture<?> existing = pendingReconciliation;
+        if (existing != null && !existing.isDone()) {
+            existing.cancel(false);
```
mfvitale, line 271: Could interrupting the provisioning lead to a bad state?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
```diff
@@ -0,0 +1,438 @@
...
+            }
+            catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                break;
+            }
+            catch (ClosedWatchServiceException e) {
+                break;
+            }
+
+            if (key == null) {
```
mfvitale: Could you just flip this so you can remove the `continue`?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
```diff
@@ -0,0 +1,438 @@
...
+
+        ReconciliationPlan plan = buildReconciliationPlan(fileHosts, dbHosts);
+
+        for (SshHostEntry entry : plan.toAdd()) {
+            repository.createPendingHost(entry.alias(), resolveHostname(entry));
+            provisioningService.provision(entry.alias());
+        }
+
+        for (String alias : plan.toRemove()) {
+            repository.markHostRemoved(alias);
```
mfvitale: What happens with the delete?

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+import io.debezium.platform.environment.host.provisioning.HostProvisioningService;
+
+/**
+ * Unit tests for the pure reconciliation logic in
+ * {@link SshConfigWatcherService#buildReconciliationPlan(List, List)}.
+ *
+ * <p>These are plain JUnit 5 tests with no {@code @QuarkusTest} — they test only
+ * the stateless comparison function, which has zero database access and zero side effects.
+ */
+class SshConfigWatcherServiceTest {
```
mfvitale: Well effectively this tests the reconciliation plan not the watcher itself. If you move the `buildReconciliationPlan` in a separate class as suggested this will be the tests for that class, then I could see another test testing the watcher itself.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
d1vyanshu-kumar: @mfvitale Right now, when a host is removed from the SSH config, we mark its status as `REMOVED` in the database. This ensures the scheduler won't deploy any new pipelines to that host going forward.

The actual pipeline undeployment (stopping and migrating active pipelines from the removed host) is planned for a future implementation - the `HostPipelineController` will handle that lifecycle. For this iteration, marking as `REMOVED` is the safe first step that prevents further deployments while the undeployment logic is built out.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
d1vyanshu-kumar: @mfvitale  no, it can't. The `cancel(false) `only cancels tasks that are still waiting in the 2-second debounce window (i.e., the reconciliation hasn't started yet). The `false` parameter means "do not interrupt if already running" - so if a reconciliation is already in progress and executing provisioning calls, it will run to completion uninterrupted.

Essentially, the cancel only prevents duplicate reconciliation runs from being queued during rapid file events - it never interrupts an active one.

mfvitale: Thanks for the explanation.

d1vyanshu-kumar:
Hi @mfvitale, thanks again for the thorough review and the guidance!

I just pushed a new commit. I've completely refactored the persistence layer to align with the Blaze-Persistence architecture. We now have a proper HostStatusService extending AbstractService, along with the HostStatus and HostStatusReference entity views.

I also took your advice on separating concerns—I pulled all the pure reconciliation math out of the Watcher and put it into a dedicated HostReconciliationPlanBuilder class. The tests are much cleaner now too, since they only test the pure logic without needing to mock the entire Watcher.

The minor points (flipping the if guard, etc.) are also sorted, and I noted your point about the REMOVED status for the future pipeline undeployment implementation.

Let me know if everything looks good to go now!

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/HostReconciliationPlanBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Pure-function builder that compares SSH config file state against database
+ * state and produces a {@link ReconciliationPlan}.
+ *
+ * <p>This class is intentionally <strong>not</strong> a CDI bean — it has no
+ * injected dependencies, no mutable state, and no side effects. Every method
+ * is a deterministic transformation of its inputs.
+ */
+class HostReconciliationPlanBuilder {
```
mfvitale: Why this is not a bean? I think it should be.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
```diff
@@ -232,39 +230,37 @@ private void watchLoop() {
...
 
-                @SuppressWarnings("unchecked")
-                WatchEvent<Path> pathEvent = (WatchEvent<Path>) event;
-                Path changedFile = pathEvent.context();
-
-                if (changedFile.equals(configFileName)) {
-                    logger.debugv("SSH config file event: {0} — {1}",
-                            event.kind().name(), changedFile);
-                    configFileAffected = true;
+                if (configFileAffected) {
```
mfvitale: Should this be inside the `if` above?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/HostReconciliationPlanBuilder.java
d1vyanshu-kumar: @mfvitale I kept it as a plain Java object instantiated with new because it's completely stateless and acts as a pure function (no dependencies or side effects).

mfvitale:
@d1vyanshu-kumar Could you also add an IT for the Watcher?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/HostReconciliationPlanBuilder.java
d1vyanshu-kumar: I think you make a good point about consistency — I can see that making it a bean keeps the dependency graph explicit and follows the same pattern as the rest of the codebase. I'll convert it to an **`@ApplicationScoped`** bean and inject it into the watcher in the next commit.

d1vyanshu-kumar:
> @d1vyanshu-kumar Could you also add an IT for the Watcher?

Sure, I'll add one! I'm thinking of a `@QuarkusTest `that writes a temp SSH config file, verifies the initial reconciliation picks it up in the database, and then modifies the file to check that the watcher detects the change and updates the DB state accordingly. I'll include it in the next commit.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
d1vyanshu-kumar: @mfvitale, I can see how this looks odd at first glance. The reason it's outside the loop is to handle the OVERFLOW case cleanly. When WatchService overflows, we set the flag and break out of the loop. The if (configFileAffected) check after the loop catches both the normal case (our config file was modified) and the overflow case (we missed events and need a full resync) in a single place.

That said, I could also just call scheduleReconciliation() directly from both places and drop the flag entirely — let me know if you'd prefer that, happy to refactor it either way.

mfvitale: > That said, I could also just call scheduleReconciliation() directly from both places and drop the flag entirely — let me know if you'd prefer that, happy to refactor it either way.

Yes, this is exactly where I was leading you.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
```diff
@@ -0,0 +1,370 @@
...
+
+        plan.toAdd().forEach(entry -> {
+            hostStatusService.createPendingHost(entry.alias(), planBuilder.resolveHostname(entry));
+            provisioningService.provision(entry.alias());
+        });
+
+        plan.toRemove().forEach(hostStatusService::markHostRemoved);
+
+        plan.toUpdate().forEach(entry -> {
+            hostStatusService.updateHostDetails(entry.alias(), entry.hostname());
```
Naros: Should this also use planBuilder.resolveHostname(entry)?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
```diff
@@ -0,0 +1,370 @@
...
+
+    /**
+     * Core reconciliation logic. Called from three places:
+     * <ol>
+     *   <li>{@link #onStart} — immediately on application startup</li>
+     *   <li>{@link #watchLoop} — via {@link #scheduleReconciliation}, after file change + debounce</li>
+     *   <li>{@link #scheduledReconciliation} — every 5 minutes as fallback</li>
+     * </ol>
+     */
+    private void reconcile() {
```
Naros: Given that this method can be called from three concurrent sources, should this method be synchronized and/or a ReentrantLock used to prevent concurrent reconciliation from happening?

Naros (commented):
Hi @mfvitale @d1vyanshu-kumar I've left a few questions below if you could take a quick look for me.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
d1vyanshu-kumar: @Naros  Good catch, thank you! Yes, you're right - the `toUpdate` block should use `planBuilder.resolveHostname(entry) `for consistency with `toAdd`. Without it, entries that omit the `HostName` directive would resolve to `null` instead of falling back to the alias. I'll fix this.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherService.java
d1vyanshu-kumar: That's a valid point - the debounce thread and the scheduled fallback could overlap if a file change happens right at the 5-minute mark. I'll add `synchronized` to the `reconcile() `method to prevent concurrent execution. A `ReentrantLock` would work too, but `synchronized` keeps it simpler since we don't need try-lock or condition semantics here. Thanks for flagging this!

d1vyanshu-kumar:
> Hi @mfvitale @d1vyanshu-kumar I've left a few questions below if you could take a quick look for me.

Hi @Naros, thank you so much for the review and for catching this! It's definitely a genuine bug. I will wait for @mfvitale to take a look and share his POV, but I am fully ready to push the fix

Naros (approved):
LGTM, @mfvitale will be back next week, waiting for his input.

mfvitale:
@d1vyanshu-kumar Applied! Thanks!

PR #461, opened 2026-07-09
debezium/dbz#2212 Add support to toggle to glass contrast and make it default for the dark mode

indraraj:
Fixes debezium/dbz#2212

## Description
Add support for glass contrast and a contrast toggle 

Screenshot

https://github.com/user-attachments/assets/556fe9e6-e6df-4ef9-b523-82b2bb0042be

## PR Checklist


mfvitale:
@indraraj Honestly I'm not a huge fan of `glass` effect, I strongly prefer the `flat` style. That said we can keep but I will just not set it has default. WDYT?

indraraj:
I primarily added it for the dark theme to improve visibility. 
It only set as the default for the dark theme. For the light theme, the appearance remains the same as it is currently as glass contrast is disabled by default there. 

In the video, if you see when I toggle to dark mode, it also applies glass contrast. And as soon as I move to the light theme, it removes the glass contrast and added only when I manually enable it.

WDYT, Fine as a default for dark, or to keep the flat as the default in dark theme too? @mfvitale

mfvitale:
> I primarily added it for the dark theme to improve visibility. It only set as the default for the dark theme. For the light theme, the appearance remains the same as it is currently as glass contrast is disabled by default there.
> 
> In the video, if you see when I toggle to dark mode, it also applies glass contrast. And as soon as I move to the light theme, it removes the glass contrast and added only when I manually enable it.
> 
> WDYT, Fine as a default for dark, or to keep the flat as the default in dark theme too? @mfvitale

For me 100% on full flat for dark and light but I would like to have a feedback also from the others. 

@Naros @vsantonastaso @kmos WDYT?

vsantonastaso:
> > I primarily added it for the dark theme to improve visibility. It only set as the default for the dark theme. For the light theme, the appearance remains the same as it is currently as glass contrast is disabled by default there.
> > In the video, if you see when I toggle to dark mode, it also applies glass contrast. And as soon as I move to the light theme, it removes the glass contrast and added only when I manually enable it.
> > WDYT, Fine as a default for dark, or to keep the flat as the default in dark theme too? @mfvitale
> 
> For me 100% on full flat for dark and light but I would like to have a feedback also from the others.
> 
> @Naros @vsantonastaso @kmos WDYT?


I like the glass effect because it makes everything look more three-dimensional and modern.
So, if I had to choose, I'd keep it as the default setting for both themes. But it's not a strong preference.

Naros:
So I do think the `glass` mode will definitely have some appeal to certain users. 

However, I'm aligned with Mario about it not being the default. The main reason is real estate.  The `glass` effect adds more space around the left navigation bar, reducing the render area where the user interacts most. For wide screens, this is likely a non-issue, but on smaller screens, it could be problematic.

So I generally would err on the side of maximizing real estate as the default, and being fancier for those special niche cases for users who have the screen real estate to afford it.

rk3rn3r:
Probably one of the reasons for the negative feedback could be switching between glass and flat when switching between light and dark mode. The layout moves a lot then, making the change more dramatic than it is.
Probably it's the same with going from clean white theme in flat mode to the blue-ish glass design. wdyt?

I prefer the more professional looking, very clean flat design, but I have to admit it might be because of the above reasons.

indraraj:
Thanks for the feedback. I agree that the glass contrast uses more padding between elements in turn also causing more layout movement.
I will keep the Glass Contrast option and lt up to the user to enable it. By default UI will be in flat contrast for both light and dark modes.

PR #463, opened 2026-07-16
debezium/dbz#1597 Add support to download the dbz server configurion for a pipeline in UI

### LINKED ISSUE #1597: Debezium Platform: Add support to generate the Debezium server config from platform pipeline.

_Add support use the platform UI also to generate the Debezium server config from the data pipeline configured in the debezium platform_

`Currently, we already support directly importing a Debezium Server configuration into the Platform UI to create a data pipeline. 
We can extend this by allowing users to take an existing pipeline configured in the Debezium Platform and deploy it on bare metal, by providing an option either in the final step of pipeline creation or afterwards to export a Debezium Server configuration that can be used directly with Debezium Server.
`

indraraj:
Fixes debezium/dbz#1597

## Description
Allow the user to download the debezium serve config file for any running pipeline in the Platform.
Redact any password|secret|credential

Screenshot

https://github.com/user-attachments/assets/64cfdfcd-7773-4ca3-8483-7f913cbb5b63

Example content of downloaded file
image

## PR Checklist


indraraj:
@mfvitale let me know wdyt?

mfvitale:
@indraraj I think it's a good addition! LGMT

I wonder if we could think to add a "multi-select" mode on the pipeline list so for example this kind of operation could be done for the selected pipeline providing a zip file? Just an idea for a future improvement.

indraraj:
Sure, I'll add a multi-select option.

Do you think we should also provide an "Extract Server Configuration" option as part of the pipeline creation flow? So that if users want they can extract the server configuration just before creating the actual pipeline

On predicates, is the following a valid Debezium Server configuration? here by default, a predicate name is added for each transform and if the predicate is not configured its type is set to `null`.
e.g
```
debezium.predicates=header_2_valPredicate,Debezium_markerPredicate
debezium.predicates.header_2_valPredicate.type=null
debezium.predicates.Debezium_markerPredicate.type=org.apache.kafka.connect.transforms.predicates.TopicNameMatches
```

Also for the log level properties configured on the pipeline I'm currently dropping them from the generated server configuration. Is that fine or should we handle them in some other way?
@mfvitale

mfvitale:
> Do you think we should also provide an "Extract Server Configuration" option as part of the pipeline creation flow? 

I think no. 

> if the predicate is not configured its type is set to null.

It should be removed. 

> Also for the log level properties configured on the pipeline I'm currently dropping them from the generated server configuration. Is that fine or should we handle them in some other way?

I think it's fine to leave it out now.

indraraj:
> > if the predicate is not configured its type is set to null.
> 
> It should be removed.

This is because the predicate object is always present in the transform call, and when no predicate is configured, it returns an empty object.
I think the better approach would be to fix this in the backend by making the predicate field optional and omitting it from the response when its not configured.

But if this current output is an invalid config, I can add a fix in the UI to handle it.

mfvitale:
> This is because the predicate object is always present in the transform call,

What is the REST API are you referring to?

indraraj:
> > This is because the predicate object is always present in the transform call,
> 
> What is the REST API are you referring to?

`/transforms` `GET`

In this, even with no predicate configured, it always returns the predicate object 

`{"id":14,"name":"Debezium marker without pridicate","description":"Extract Debezium payload","type":"io.debezium.transforms.ExtractNewRecordState","schema":"string","vaults":[],"config":{"add.fields":"op","add.headers":"db,table"},"predicate":{"type":null,"config":null,"negate":false}}`

mfvitale:
@indraraj Could you open an issue. I think the return should be `predicate: {}` in case of no predicates specified.

indraraj:
Sure, do you think in the meantime `debezium.predicates.{transform_name}.type=null` is valid, or I need to rectify it in UI? @mfvitale

mfvitale:
> Sure, do you think in the meantime `debezium.predicates.{transform_name}.type=null` is valid, or I need to rectify it in UI? @mfvitale

No it is not valid. So if you see `type` null from the response, you should remove it.

indraraj:
Updated to remove the predicate object if the `type` defined is set to `null`, Backend issue https://github.com/debezium/dbz/issues/2308

PR #476, opened 2026-07-31
debezium/dbz#2261 Add the Pipeline failure status, error msg and ability to restart

indraraj:
Fixes debezium/dbz#2261

## Description
- Show the Pipeline status (Currently only Failed)
- Failuer erorr msg
- Ability to restart a pipeline

Screenshot

https://github.com/user-attachments/assets/fb092df2-5531-44b8-bfda-58c8f125a243

## PR Checklist

### DIFF
--- a/debezium-platform-stage/cypress/e2e/pipeline.cy.ts
+++ b/debezium-platform-stage/cypress/e2e/pipeline.cy.ts
@@ -435,6 +435,107 @@ describe('Pipeline Management', () => {
       cy.contains('View logs').click();
       cy.url({ timeout: 30000 }).should('match', /\/pipeline\/\d+\/logs/);
     });
+
+    it('should show FAILED status cell with danger icon and link in pipeline list', () => {
+      // Seed a pipeline via API then stub the list response to mark it FAILED
+      const pipelineName = `cypress-failed-status-${Date.now()}`;
+      createPipelineViaApi(pipelineName);
+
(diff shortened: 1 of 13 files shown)


mfvitale:
@indraraj Could we add just a more general message instead of the `errorMessage` from the API. Maybe we can add a generic one and the possibility to see the detailed one?

indraraj:
Something like this, with generic msg being "Failed to start the Pipeline" @mfvitale 



https://github.com/user-attachments/assets/fdfa7242-745d-4b9f-bf0a-a7f8da5f8cb3

indraraj:
For the failed pipeline, I am disabling the `Action` and `Monitoring` tabs in UI. WDYT about logs, Will we still have logs available in this case?
Also, I noticed that after triggering a restart the `Failed` status remains unchanged in the subsequent calls instead of following the expected behaviour of clearing it immediately and showing it again only if the restart would fail. Thus from UI currently it feels like the restart button isn't doing anything.

mfvitale:
> For the failed pipeline, I am disabling the `Action` and `Monitoring` tabs in UI. WDYT about logs, Will we still have logs available in this case?

Yes.

> Also, I noticed that after triggering a restart the Failed status remains unchanged in the subsequent calls instead of following the expected behaviour of clearing it immediately and showing it again only if the restart would fail. Thus from UI currently it feels like the restart button isn't doing anything.

Yes, for now we only manage the `FAIL` status. I opened and issue https://github.com/debezium/dbz/issues/2351

PR #309, opened 2026-03-21
debezium/dbz#1245 Add description field to connections API payload

PDGGK:
## Summary

Adds the `description` field to the connections API payload, consistent with all other resources (pipelines, sources, destinations, transforms, vaults) that already have this field.

Closes debezium/dbz#1245

## Changes

**Backend (Java/Quarkus):**
- `ConnectionEntity.java`: Added `description` field with getter/setter
- `Connection.java` (Blaze-Persistence EntityView): Added `getDescription()`/`setDescription()`
- New Flyway migration `V3.5.0__add_connection_description.sql`: Adds `description varchar(255)` column to `connection` table

**Frontend (TypeScript/React):**
- `apis.tsx`: Added `description?: string` to `Connection` and `ConnectionPayload` types
- `Connections.json`: Updated mock data with description field

**Tests:**
- `ConnectionResourceIT.java`: Updated test payloads to include description, added assertions, added `testCreateConnection_WithoutDescription` to verify optionality

## Design Decisions
- Description is optional (nullable), consistent with all other entities
- No REST endpoint changes needed — Blaze-Persistence and Jackson handle serialization automatically
- Column type `varchar(255)` matches existing description columns in other tables

PR #477, opened 2026-08-02
debezium/dbz#2250 Add the description field to the connection form, view and edit pages

PDGGK:
Fixes debezium/dbz#2250

## Description

Adds the connection `description` field to the UI. This is the frontend half of debezium/dbz#1245, whose backend and API work merged in #309.

#309 already shipped the type side — `description?: string` on both `ConnectionPayload` and `Connection` in `src/apis/apis.tsx`, plus sample values in `src/__mocks__/data/Connections.json`. This puts the UI on top of it.

**`CreateConnection.tsx`**
- an optional description `TextInput` directly below the name field, with helper text
- declared as `description: yup.string().notRequired()` in the form schema
- carried into the object built by `buildConnectionPayloadOrNull`, in both the schema and the no-schema branch

**`EditConnection.tsx`**
- the saved description is seeded from the fetched connection in the `useForm` defaults and in **both** `reset()` branches. Without that, opening an existing connection and saving any unrelated change silently wipes its description.
- rendered read-only in view mode through the existing `ReviewValueSpan`, so a connection with no description shows the same `—` placeholder as the other fields
- same payload wiring on the PUT

**Locales** — `connection:form.descriptionField` and `connection:form.descriptionFieldHelperText`, added to both `en` and `it`.

### The second commit: resetting scroll on navigation

The first push of this branch turned two long-standing Cypress assertions red — `should display create connection form for PostgreSQL` and `… for Kafka destination`, both at `cy.contains('h1', 'Create connection').should('be.visible')`. I reproduced it locally against the compose stack and measured it rather than guessing, and it is worth explaining because the cause is not the field itself.

`<main id="primary-app-container">` is the app's scroll container, and a client-side route change leaves its `scrollTop` untouched. In those two tests Cypress scrolls the catalog to reach the connector card, clicks it, and React Router swaps in the create page with that offset still applied. The offset only survives if the new page is tall enough to keep it — otherwise the browser clamps it away.

Measured on the create-connection page immediately after that navigation:

| | `scrollTop` | max scroll | `<h1>` top |
| --- | --- | --- | --- |
| before this branch | 48.5 | 624 | **78** (container starts at 73 — visible by ~5px) |
| with the description field | 160.5 | 736 | **−34** (clipped) |
| with the scroll reset | 0 | 736 | **126** (same as a direct visit) |

So the assertions were passing with about five pixels to spare, and one extra form row was enough to tip them over. The test is right — landing on a page with its heading above the fold is a real defect — so the second commit resets `scrollTop` on `location.pathname` change, next to the existing pathname-keyed effect in `AppLayout`. It is five lines and needs no new imports.

I have kept it as a separate commit since it is a distinct concern; happy to split it into its own PR if you would rather review it on its own, though this branch does need it to stay green.

### A few choices worth calling out

- **Which namespace the labels live in.** `CreateSchemaForm` keeps its description strings in `common` because it is shared between source and destination. `pipeline` and `transform` keep theirs in their own namespace. A connection-only form falls in the second group, so the keys are in `connection.json`.
- **Blank description.** The payload always carries `description`, sending `""` when the field is empty, which is what `CreateSchemaForm.tsx` does. That makes clearing a description behave the same way regardless of how the mapper treats an absent field, and `reviewValue()` already renders `""` as `—`.
- **`/api/connections/validate`** binds the same `ConnectionRequest` record as create and update, so the extra field rides along harmlessly on validation requests.
- **No description column in the connections list.** Sources and destinations have had a description for a while and `SourceSinkTable` still does not show it, so I left `ConnectionTable` alone for consistency. Happy to add it if you would rather have it.
- **`ServerConfigModal` left alone.** It is the other place that POSTs to `/api/connections`, but it auto-provisions connections with faker-generated names while importing a Debezium Server config, so there is no user-entered description to carry. Easy to add a canned one there later if that would be useful.

## Testing

New `CreateConnection.test.tsx` (4 tests) and `EditConnection.test.tsx` (7 tests). All eleven fail when the two source files are reverted to `main`, so they are genuine regression guards rather than tests that pass by construction — the edit-page ones specifically cover the wipe-on-save case described above.

Both components build the payload in two branches, depending on whether a schema was resolved for the connection type. Since every connector in `connection-schemas.json` takes the schema-backed branch, each component has a test that drives it end to end — Validate, then save — and asserts both that `description` reaches the payload and that it does **not** appear in `config`. Removing only the schema-branch `description` line fails exactly those two tests and nothing else.

Local run of the `stage-workflow` chain:

| step | result |
| --- | --- |
| `yarn lint` | 0 errors (the 35 warnings are pre-existing; none in `src/pages/Connection`) |
| `yarn type-check` | clean |
| `yarn build` | ok |
| `yarn test` | 70 files / 402 tests pass, up from 68 / 391 |
| `yarn cypress run` | all 6 specs, 134 / 134 pass against the compose stack |

The Cypress run is the full suite, not just `connection.cy.ts`, because the scroll change touches every page.

## Two pre-existing things I noticed but did not touch

- `connection:form.nameExists` is referenced in both files but is missing from both locale files, so it only ever renders its inline English default. The new `form` block would be its natural home — say the word and I will add it here or in a separate PR.
- In `EditConnection`'s fetch effect, neither `reset()` branch runs when a connector's type matches a schema but the API returns a falsy `config`. On that path `name` is left blank too, and `name` is `required()`, so the form simply refuses to submit rather than saving anything away — the visible symptom is view mode showing `—`. Pre-existing and not specific to this change.
- The single-input form field name is `description`, matching `name` alongside it. As with `name`, a connector schema that declared a config property of the same literal name would shadow it, since `buildNestedConnectionYupFields` is spread after the explicit entries. None of the sixteen schemas in `connection-schemas.json` declares either, so this is theoretical today — mentioning it only because I checked.

## PR Checklist

### DIFF
--- a/debezium-platform-stage/src/pages/Connection/CreateConnection.tsx
+++ b/debezium-platform-stage/src/pages/Connection/CreateConnection.tsx
@@ -1,5 +1,5 @@
 import PageHeader from "@components/PageHeader";
-import { ActionList, ActionListGroup, ActionListItem, Alert, Button, Card, CardBody, Content, Form, FormAlert, FormFieldGroup, FormFieldGroupHeader, FormGroup, FormGroupLabelHelp, FormHelperText, HelperText, HelperTextItem, InputGroup, InputGroupItem, PageSection, Popover, TextInput } from "@patternfly/react-core";
+import { ActionList, ActionListGroup, ActionListItem, Alert, Button, Card, CardBody, Content, Form, FormAlert, FormFieldGroup, FormFieldGroupHeader, FormGroup, FormGroupLabelHelp, FormHelperText, HelperText, HelperTextItem, InputGroup, InputGroupItem, PageSection, Popover, Skeleton, TextInput } from "@patternfly/react-core";
 import * as React from "react";
 import _, { } from "lodash";
 import { Controller, useForm } from "react-hook-form";
@@ -61,7 +61,7 @@ const CreateConnection: React.FunctionComponent<ICreateConnectionProps> = ({ sel
     const [properties, setProperties] = useState<Map<string, AdditionalPropertyRow>>(() => new Map([["key0", createEmptyAdditionalPropertyRow()]]));
     const [keyCount, setKeyCount] = React.useState<number>(1);
 
-    const { data: connectionsSchema = [] } = useQuery<ConnectionsSchema[], Error>("connectionsSchema", () =>
+    const { data: connectionsSchema = [], isLoading: isSchemaLoading } = useQuery<ConnectionsSchema[], Error>("connectionsSchema", () =>
         fetchData<ConnectionsSchema[]>(`${API_URL}/api/connections/schemas`)
     );
 
@@ -303,6 +303,17 @@ const CreateConnection: React.FunctionComponent<ICreateConnectionProps> = ({ sel
 
                 <Card className="custom-card-body">
                     <CardBody isFilled>
+                        {isSchemaLoading ? (
+                            <div>
+                                <Skeleton fontSize="2xl" width="40%" />
+                                <br />
+                                <Skeleton fontSize="md" width="60%" />
+                                <br />
+                                <Skeleton fontSize="md" width="80%" />
+                                <br />
+                                <Skeleton fontSize="md" width="50%" />
+                            </div>
+                        ) : (
                         <Form id="create-connection-form" onSubmit={handleSubmit(handleSaveFromForm)} isWidthLimited>
                             {!_.isEmpty(errors) && (
                                 <FormAlert>
@@ -512,6 +523,7 @@ const CreateConnection: React.FunctionComponent<ICreateConnectionProps> = ({ sel
                                 </>
                             </FormFieldGroup>
                         </Form>
+                        )}
                     </CardBody>
                 </Card>
 
--- a/debezium-platform-stage/src/pages/Connection/CreateConnection.tsx
+++ b/debezium-platform-stage/src/pages/Connection/CreateConnection.tsx
@@ -103,6 +103,7 @@ const CreateConnection: React.FunctionComponent<ICreateConnectionProps> = ({ sel
                 return !conns.some(c => c.name === value);
             }
         ),
+        description: yup.string().notRequired(),
         ...buildNestedConnectionYupFields(selectedSchemaProperties)
     }).required() as yup.ObjectSchema<ConnectionFormValues>;
 
@@ -218,7 +219,7 @@ const CreateConnection: React.FunctionComponent<ICreateConnectionProps> = ({ sel
     };
 
     const buildConnectionPayloadOrNull = (data: ConnectionFormValues): ConnectionPayload | null => {
-        const { name } = data;
+        const { name, description } = data;
         const schemaPropertyKeys = selectedSchemaProperties?.properties
             ? Object.keys(selectedSchemaProperties.properties)
             : [];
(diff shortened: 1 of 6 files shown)


PDGGK:
The first push turned two existing Cypress assertions red, so I chased that down before asking anyone to look at this. It was mine, but the cause was not the field — writing it up here since the PR description only has the summary.

`should display create connection form for PostgreSQL` and `… for Kafka destination` both fail at `cy.contains('h1', 'Create connection').should('be.visible')`. I confirmed the branch was responsible (27/27 with the two source files reverted to `main`, 25/27 with them), reproduced it locally against the compose stack, and then measured it with a throwaway spec that dumped `getBoundingClientRect()` and every scrollable ancestor rather than guessing from the screenshot.

`<main id="primary-app-container">` is the app's scroll container and its `scrollTop` survives a client-side route change. In those two tests Cypress scrolls the catalog down to reach the connector card, clicks it, and React Router swaps in the create page with that offset still applied — it only disappears if the destination page is short enough for the browser to clamp it away.

On the create-connection page immediately after that navigation:

| | `scrollTop` | max scroll | `<h1>` top |
| --- | --- | --- | --- |
| `main` | 48.5 | 624 | 78 (container starts at 73) |
| this branch, before the fix | 160.5 | 736 | −34 |
| this branch, with the fix | 0 | 736 | 126 |

So the assertions were clearing the container's top edge by about five pixels, and one extra form row was enough to tip them over. Any field added to that form next would have done the same.

I did consider just dropping the helper text to save the height, but that hides the defect rather than fixing it, and the assertion is describing correct behaviour — landing on a page with its heading above the fold is a real bug. So `c6cb1d8` resets `scrollTop` on `location.pathname` change, right next to the existing pathname-keyed effect in `AppLayout`. Five lines, no new imports.

Since that commit affects every page, I ran the whole Cypress suite locally rather than only the connection spec: 6 specs, 134/134. Happy to pull it into its own PR if you would prefer to review it separately — this branch just needs it to stay green.

debezium-platform-stage/src/appLayout/AppLayout.tsx
```diff
@@ -101,6 +101,16 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({ children }) => {
     }
   }, [location.pathname, setSelectedSource, setSelectedDestination, setSelectedTransform]);
 
+  // A client-side route change keeps the scroll offset of the page we came from,
```
indraraj: I think let's avoid this fix from here, instead we can prevent the scroll issue on the Create Connection page by adding an on-load Skeleton before we update the page with an actual form-based schema.

debezium-platform-stage/src/pages/Connection/CreateConnection.tsx
indraraj, line 305: You can add a preload Skeleton check here 
```
isSchemaLoading ? (
    <div>
      <Skeleton fontSize="2xl" width="40%" />
       <br />
       <Skeleton fontSize="md" width="60%" />
       <br />
       <Skeleton fontSize="md" width="80%" />
       <br />
       <Skeleton fontSize="md" width="50%" />
   </div>
```

indraraj:
@PDGGK tested locally, looks good. Just added one suggestion

PDGGK:
Thanks @indraraj — done in `39fa7e4`. The `AppLayout` change is dropped and the skeleton is in `CreateConnection` exactly where you pointed, using the shape from your suggestion.

I verified it rather than assuming, because the fix works through a slightly indirect mechanism and I wanted to be sure it holds:

| | main scroll offset on arrival | `<h1>` top |
| --- | --- | --- |
| before this branch | 48.5 | 78 (container top is 73 — visible by ~5px) |
| with the description field, no skeleton | 160.5 | −34, clipped |
| with the skeleton | 0 | 126 |

The skeleton keeps the card short while `connectionsSchema` is loading, so the browser clamps the offset carried over from the catalog down to zero; by the time the real form expands the page, it is already at the top. Nice side effect: this also removes the flash of an empty form before the schema fields appear.

Cypress against the compose stack:

- `connection.cy.ts` — 27/27
- removing just the skeleton again — back to 25/27, failing on the same two `cy.contains('h1', 'Create connection').should('be.visible')` assertions
- full suite — 6 specs, 134/134

Also green: `yarn lint` (0 errors), `yarn type-check`, `yarn test` (70 files / 402 tests).

Worth noting those two assertions were only clearing the container's top edge by about five pixels before any of this, so the next field added to that form would have tipped them over the same way. The skeleton removes that fragility for good, which the `AppLayout` change would not have done as cleanly.

indraraj (approved):

indraraj:
Applied, thanks @PDGGK

PR #478, opened 2026-08-03
debezium/dbz#2308 Stop reporting an empty predicate on the transforms API

PDGGK:
Fixes debezium/dbz#2308

## Description

`GET` and `POST /transforms` report a predicate for transforms that have none:

```json
{"id":14,"name":"...","predicate":{"type":null,"config":null,"negate":false}}
```

`TransformEntity.Predicate` is an `@Embeddable`, so Hibernate materialises it whether or not a predicate was configured, leaving every field unset. The generated mapper only null-checks the reference, never its contents, so that empty instance is serialised as though it were a real predicate.

The write path already guards this — `TransformMapper.applyToView` only touches the predicate when the request carries one — and the operator side already has the right emptiness test, at `environment/operator/PipelineMapper.java:318`:

```java
return transform.getPredicate() != null && transform.getPredicate().getType() != null;
```

This applies the same test on the read path: `predicateToDto` returns `null` for a predicate with no type, and the response field is annotated `@JsonInclude(NON_NULL)` so the key is omitted rather than serialised as `"predicate": null`.

### On the wire shape

The issue asks for `predicate: {}` or no `predicate` object at all; I went with omitting the key. Worth flagging two things so you can redirect cheaply if you would rather have `{}`:

- The mapper change alone is not enough — it leaves the key present with a `null` value, because this project has no global Jackson inclusion setting and no other `@JsonInclude` usage today. The annotation is what actually removes the key.
- It is deliberately on the record component and not on the type. Type-level `NON_NULL` would silently drop `description`, `schema` and every other unset field from every transform response, which is a much bigger change than this issue is asking for.

Existing consumers are unaffected either way: `apis.tsx` already declares `predicate?`, and every frontend read of it is a truthiness check.

## Testing

New `TransformResourceIT` — `SourceResourceIT`-style `@QuarkusTest` + RestAssured, no operator or k8s harness needed:

- a transform created without a predicate has no `predicate` key, on both the create response and a subsequent `GET`
- a transform created with a predicate round-trips type, config and negate unchanged

Both pass with the change (2/2). With `TransformMapper` and `TransformResponse` reverted to `main`, the first test fails on the missing-key assertion and the round-trip test still passes, so the new coverage is pinned to this behaviour rather than passing by construction.

`process-sources checkstyle:checkstyle` with the formatter and import-order validation that CI runs is clean.

One note on my local run: the failsafe pass also tries to start `ghcr.io/project-zot/zot-linux-amd64`, which has no arm64 image, so on Apple Silicon that dev service fails before the test body executes. The results above come from the surefire pass, which brings up the same Postgres dev service and runs the same class. This does not affect CI on `ubuntu-latest`.

## PR Checklist

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/api/dto/TransformResponse.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/api/dto/TransformResponse.java
@@ -8,6 +8,8 @@ package io.debezium.platform.api.dto;
 import java.util.Map;
 import java.util.Set;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+
 public record TransformResponse(
         Long id,
         String name,
@@ -16,5 +18,5 @@ public record TransformResponse(
         String schema,
         Set<NamedRef> vaults,
         Map<String, Object> config,
-        PredicateDto predicate) {
+        @JsonInclude(JsonInclude.Include.NON_NULL) PredicateDto predicate) {
 }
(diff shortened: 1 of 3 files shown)


mfvitale (approved):
@PDGGK Applied, thanks!

PR #486, opened 2026-08-07
Dbz#2314 Alerting support to the Debezium Platform

### LINKED ISSUE #2314: Alerting support to the Debezium Platform

  The platform exposes pipeline monitoring through 14 predefined panels (DDD-38), but monitoring is passive: users must manually watch dashboards to detect problems.

  Alerting closes this gap by letting users define threshold-based rules against existing monitoring panels and receive notifications when conditions are breached.

  Key capabilities:

  - **Alert rules**: users define rules by selecting a monitoring panel, a comparison operator, a numeric threshold, and an optional "for" duration to suppress transient
  spikes
  - **Global evaluation**: rules fire for any pipeline that breaches the threshold, the alert event identifies which pipeline triggered it
  - **Alert lifecycle**: three-state model (Pending, Firing, Resolved) with incident tracking (one record per fire/resolve cycle)
  - **Severity levels**: Critical, Warning, Info
  - **Notification channels**: Email (SMTP) and Webhook (HTTP POST), plus always-on in-app visibility (history page and alert badge)
  - **Alert history**: queryable, filterable, paginated incident history with configurable retention (default 30 days)
  - **UI**: new Alerts section with Rules, Channels, and History sub-pages, plus a global alert badge in the header

  Out of scope for initial implementation: alert grouping, silencing, escalation policies, repeat notifications, raw PromQL mode, per-pipeline rule scoping, real-time push
  (SSE/WebSocket).

  ---

  **Implementation ideas (optional)**

  [DDD-57: Alerting for Debezium Platform](https://github.com/debezium/debezium-design-documents/pull/58)

mfvitale:
Fixes debezium/dbz#2314

## Description
This pull request introduces a comprehensive REST API for managing alerting and notification features in the platform conductor module. It adds new resources for alert events, alert rules, and notification channels, along with their associated request and response DTOs. The changes enable CRUD operations, status queries, and notification channel testing, and add the necessary dependencies for email notifications.

The most important changes are:

**API Resource Additions:**

* Added `AlertEventResource`, `AlertRuleResource`, and `NotificationChannelResource` classes, providing REST endpoints for querying alert events and statuses, managing alert rules (CRUD, enable/disable), and managing notification channels (CRUD, test channel) (`AlertEventResource.java` [[1]](diffhunk://#diff-f4a2d61c39e88f1522a8c5536a499a0c1e951a0040c1be3325c5e2a130c4ff3cR1-R82) `AlertRuleResource.java` [[2]](diffhunk://#diff-406e41f7bfb078e6795a561155f801eba3fd200c8bd9732d286bf115293752faR1-R125) `NotificationChannelResource.java` [[3]](diffhunk://#diff-8bb97814105e8e659a1dd8bfc14b14efe9c1b1e27c08300f081f0a07e3ac484dR1-R115).

**DTOs for API Communication:**

* Introduced DTO records for API requests and responses: `AlertEventResponse`, `AlertRuleRequest`, `AlertRuleResponse`, `AlertStatusResponse`, `NotificationChannelRequest`, `NotificationChannelResponse`, `PagedAlertEventResponse`, and `TestNotificationResponse`, standardizing data exchange formats for the new endpoints [[1]](diffhunk://#diff-32a0002d9dc29e8e8d1eaf78671141728cd13d6f6dff72fc03f78055cd3d12f8R1-R27) [[2]](diffhunk://#diff-836072ef5d86d501542ddccc3e1d6064800b00d776764b0192ff1e12e976e445R1-R32) [[3]](diffhunk://#diff-e8d9bf70a4db61d842940bd634d40fbf4ea50449ebea5eb18ba335343ed59437R1-R38) [[4]](diffhunk://#diff-c9fb278d54d8fec7019b099efff1057f1cfcc206fb04bc3c5c86babdae68e16eR1-R31) [[5]](diffhunk://#diff-4edc397f03c3e1033a29b18ba640b2300017c10c6e01c8eeb2ab4077178a883eR1-R20) [[6]](diffhunk://#diff-86e7a710c07e00e875d195adb505063d883e5fd2890b1f020ee8084f73ff7e86R1-R21) [[7]](diffhunk://#diff-eaf0cebb45c2bbf0b300a36f79be36922aa91f12c41dbc2a73d715b8ba8812a5R1-R16) [[8]](diffhunk://#diff-e84ffe69b6f379a169c6fd7564dc3fa208f7dc15cea7b300f4ba5d5b3384d944R1-R11).

**Dependency Updates:**

* Added the `quarkus-mailer` dependency to the project to support email-based notification channels.

## PR Checklist


debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertStateService.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public List<AlertStateEntity> findByRuleId(Long ruleId) {
+        return em.createNamedQuery(AlertStateEntity.FIND_BY_RULE_ID, AlertStateEntity.class)
+                .setParameter("ruleId", ruleId)
+                .getResultList();
+    }
+
+    @Transactional
+    public void evaluate(AlertRuleEntity rule, String pipelineId, double value,
```
Naros: This method and `resolve` both open a transaction that blocks until webhooks are sent synchronously, including retries that can block for ~30 seconds. Should we persist the transition and then handle the event/webhook dispatch using a CDI bean or using an `@Async` / Vert.x event bus?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertTransitionEvaluator.java
```diff
@@ -0,0 +1,55 @@
...
+@ApplicationScoped
+public class AlertTransitionEvaluator {
+
+    public StateTransition evaluate(AlertStateValue currentState, boolean conditionMet,
+                                    Duration forDuration, Instant pendingSince, Instant now) {
+        return switch (currentState) {
+            case OK -> evaluateOk(conditionMet, forDuration, now);
+            case PENDING -> evaluatePending(conditionMet, forDuration, pendingSince, now);
+            case FIRING -> evaluateFiring(conditionMet);
+            default -> new StateTransition(currentState, StateTransition.Action.NONE, null, null);
```
Naros: Should this include an explicit case for `RESOLVED`, with the default as an error case if new enum values are added but not handled here?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertRuleService.java
```diff
@@ -0,0 +1,86 @@
...
+    public AlertRuleService(EntityManager em, CriteriaBuilderFactory cbf, EntityViewManager evm,
+                            PanelConfigLoader panelConfigLoader) {
+        super(AlertRuleEntity.class, AlertRule.class, AlertRuleReference.class, em, cbf, evm);
+        this.panelConfigLoader = panelConfigLoader;
+    }
+
+    @Override
+    public AlertRule create(@Valid AlertRule view) {
+        validatePanelExists(view.getPanelId());
+        validateForDuration(view.getForDuration());
```
Naros, line 38: Should there also be validation for the `evaluationWindow` that carries the same requirement as the `forDuration` property? Same question for `update` method in this class.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertEvaluationEngine.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    private void evaluateRule(AlertRuleEntity rule, Instant now) {
+        PanelConfig panel = panelConfigLoader.loadPanels().stream()
+                .filter(p -> p.id().equals(rule.getPanelId()))
+                .findFirst()
+                .orElse(null);
+
+        if (panel == null) {
```
Naros, line 80: Does this eventually lead to stale/orphaned `AlertStateEntity` with no way to resolve them? If so, should we consider resolving them after a configurable no-data grace period?

Naros (commented):
Hi @mfvitale, this looks really clean. 🚀  I have a few questions below.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertTransitionEvaluator.java
mfvitale: @Naros This effectively works on the `currentStatus` so the case the `RESOLVED` state is never an `input` state. 

So I just removed the `RESOLVED` state and simplified the `switch`. Thanks for spotting it.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertEvaluationEngine.java
mfvitale: Good catch! I'll just auto-resolve in that case and so if a panel is then re-added and the alters is still firing it will be correctly re-evaluated. 

Note that for now the panels, are just configured via k8s and not via UI so adding/removing will be less frequent.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertStateService.java
mfvitale: Good point. Effectively this is the "dual-write" problem resolvable with the "outbox` pattern :D 

For now I think that `CDI events` is reasonable solution for this uses case.

mfvitale:
@Naros Could I ask another pass?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/notifications/WebhookNotifier.java
```diff
@@ -0,0 +1,171 @@
...
+                Thread.currentThread().interrupt();
+                return new NotificationResult(false, "Interrupted during webhook delivery");
+            }
+            catch (Exception e) {
+                LOGGER.warnv("Webhook failed (attempt {0}/{1}): {2}", attempt + 1, maxAttempts, e.getMessage());
+            }
+
+            if (attempt < maxAttempts - 1) {
+                try {
+                    Thread.sleep(RETRY_DELAYS_MS[attempt]);
```
Naros: With max-attempts being greater than 4, this yields an ArrayIndexOutOfBoundsException.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/notifications/WebhookNotifier.java
```diff
@@ -0,0 +1,171 @@
...
+        String payload;
+        try {
+            payload = buildPayload(notification);
+        }
+        catch (JsonProcessingException e) {
+            return new NotificationResult(false, "Failed to serialize payload: " + e.getMessage());
+        }
+
+        int maxAttempts = webhookConfig.maxAttempts();
+        for (int attempt = 0; attempt < maxAttempts; attempt++) {
```
Naros: This looks similar to RetryingCallable and related concepts in the Debezium Utils module. I wonder whether we're importing that here as a dependency, and if not, would it make sense to encapsulate the same infrastructure here if retry mechanisms are needed elsewhere in the Platform?

debezium-platform-conductor/src/test/java/io/debezium/platform/domain/alert/AlertEvaluationEngineTest.java
```diff
@@ -0,0 +1,297 @@
...
+                new PrometheusInstantQueryResponse.Data("vector", List.of(
+                        new PrometheusInstantQueryResponse.Result(
+                                Map.of("other_label", "value"),
+                                List.of(1234567890.0, "42.5")))));
+
+        assertThat(engine.extractByPipeline(response)).isEmpty();
+    }
+
+    @Test
+    void extractByPipelineSkipsNonNumericValue() {
```
Naros, line 156: So if the rule is a NOT_EQUAL rule, a NaN would evaluate to true, so it won't be skipped as this method expects. Should NaN simply not be allowed as part of evaluation and retained in `extractByPipeline`?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertEventService.java
```diff
@@ -0,0 +1,179 @@
...
+        TypedQuery<Long> countQuery = em.createQuery(
+                "SELECT COUNT(e) FROM alert_event e" + whereClause, Long.class);
+        TypedQuery<AlertEventEntity> dataQuery = em.createQuery(
+                "SELECT e FROM alert_event e" + whereClause + " ORDER BY e.createdAt DESC", AlertEventEntity.class);
+
+        bindParameters(countQuery, severity, pipelineId, ruleId, from, to);
+        bindParameters(dataQuery, severity, pipelineId, ruleId, from, to);
+
+        long totalElements = countQuery.getSingleResult();
+        int totalPages = (int) Math.ceil((double) totalElements / size);
```
Naros: Here `size` is capped at 100, but not floored, so a size of 0 yields division by zero. Should this be clamped as `[1,100]`?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertEventService.java
```diff
@@ -0,0 +1,179 @@
...
+
+        dataQuery.setFirstResult(page * size);
```
Naros: Same here for the page being negative, right?

helm/README.md
```diff
@@ -128,6 +128,71 @@ The following operators must be installed in the cluster **before** deploying th
...
 | monitoring.prometheus.serviceMonitor.scrapeInterval | Prometheus scrape interval                                                                                                                                                            | 15s                                        |
 | monitoring.prometheus.serviceMonitor.labels | Labels for Prometheus Operator ServiceMonitor discovery                                                                                                                               | {prometheus: kube-prometheus}              |
+| monitoring.prometheus.serviceMonitor.scrapeInterval | Prometheus scrape interval                                                                                                                                                    | 15s                                        |
+| monitoring.prometheus.serviceMonitor.labels | Labels for Prometheus Operator ServiceMonitor discovery                                                                                                                                | {prometheus: kube-prometheus}              |
```
Naros: These seem to be duplicated?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertEvaluationEngine.java
```diff
@@ -0,0 +1,190 @@
...
+        LOGGER.debugv("Starting evaluation cycle for {0} enabled rule(s)", rules.size());
+        Instant now = Instant.now();
+
+        for (AlertRuleEntity rule : rules) {
+            try {
+                evaluateRule(rule, now);
+            }
+            catch (Exception e) {
+                LOGGER.errorv(e, "Failed to evaluate rule ''{0}''", rule.getName());
+            }
```
Naros, line 70: I missed this before, but during evaluateRule, if an exception occurs, we let the transaction the AlertStateManager joined keep iterating. Depending on the driver and database, I'm concerned the transaction gets marked for rollback only (aka aborted), and any future attempt to mutate any state results in a SQL failure. Is that intended?

Naros (commented):
Hi @mfvitale I did another pass and found a few more things when looking a bit closer. Let me know what you think.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/notifications/WebhookNotifier.java
mfvitale: I'll use it.

debezium-platform-conductor/src/test/java/io/debezium/platform/domain/alert/AlertEvaluationEngineTest.java
mfvitale: Are you sure?

I added the following test

```java
 @Test
    void testNaNValues() {
        assertThat(Operator.LESS_THAN.evaluate( Double.NaN, -3.0)).isFalse();
        assertThat(Operator.GREATER_THAN.evaluate(Double.NaN, 5.0)).isFalse();
        assertThat(Operator.EQUAL.evaluate(Double.NaN, 10.0)).isFalse();
    }
```

and it passes.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertEventService.java
mfvitale: Good catch. I added a validation for `page` and `size` at the rest level

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertEvaluationEngine.java
mfvitale: Good catch. It was not intended. So I removed it from the `evaluateAll`

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/notifications/WebhookNotifier.java
mfvitale: @Naros done.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertEvaluationEngine.java
```diff
@@ -0,0 +1,189 @@
...
+    String removeServiceNameFilter(String panelQuery) {
+        return panelQuery
+                .replace("service_name=\"{{pipeline_id}}\",", "")
+                .replace(",service_name=\"{{pipeline_id}}\"", "")
+                .replace("service_name=\"{{pipeline_id}}\"", "");
+    }
+
+    private String aggregateByPipeline(String query) {
+        return "sum by (service_name) (" + query + ")";
+    }
```
Naros, line 166: So in `panels.yml`, the `service_name` is already aggregated away, right? So this yields no service names, and every result is therefore skipped. Worse, the rule appears configured properly but never fires in `AlertRuleService#validatePanelExists`. Am I right?

Should we warn when a response has no results but none carry a service name, or reject such panels at rule-creation time? I don't think we have any tests that cover this use case, so maybe it's not valid?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/notifications/WebhookNotifier.java
```diff
@@ -0,0 +1,180 @@
...
+                for (InetAddress addr : addresses) {
+                    if (addr.isLoopbackAddress() || addr.isSiteLocalAddress()
+                            || addr.isLinkLocalAddress() || addr.isAnyLocalAddress()) {
+                        return new NotificationResult(false,
+                                "Webhook URL resolves to non-public address: " + addr.getHostAddress());
+                    }
+                }
+            }
+        }
+        catch (UnknownHostException e) {
```
Naros, line 149: Should we also catch `IllegalArgumentException` here?

Consider a URL as `http://exa mple.com` or some other typo. In this case, the validate config pass only verifies that the value is non-blank, but here we only check for an unknown host, not for any other failure condition with the URL string. WDYT?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertEventService.java
```diff
@@ -0,0 +1,140 @@
...
+        return findById(id).map(this::toResponse);
+    }
+
+    @Transactional
+    public AlertEventEntity createFiringEvent(AlertRuleEntity rule, String pipelineId,
+                                              Double value, String message, Instant firedAt) {
+        AlertEventEntity event = new AlertEventEntity();
+        event.setRule(rule);
+        event.setRuleName(rule.getName());
+        event.setPipelineId(pipelineId);
```
Naros: So we never set `pipelineName` here, and the email body/webhook payloads always use `pipelineId`. Should we populate that field somewhere, or should we drop the name? We have only ever seen it used in tests.

Naros (commented):
Hi @mfvitale a few more findings I see, let me know what you think.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/alert/AlertEvaluationEngine.java
mfvitale: Well, in the `panels.yml` the `service_name` is parameterized using the `pipelineId` since them are used in the pipeline monitoring tab to show metrics from a specific pipeline. For the alerting, since they are global, we [remove the `service_name`](service_name), and then just aggregate again by service name to see what pipeline has effectively triggering an alert. @Naros is it more clear?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/notifications/WebhookNotifier.java
mfvitale: `InetAddress.getAllByName("http://exa mple.com");` will still throw the `UnknownHostException`

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/AlertEventService.java
mfvitale: This is a good catch @Naros . Effectively here the `pipelineId` is the pipeline `name`. I checked that this wrong naming was present also in the monitoring part and since the FE is already expecting `pipelineId` (is effectively passes the name of the pipeline), I'll plan to correct them in a [follow-up](https://github.com/debezium/dbz/issues/2529) pr. Are you ok?

Naros (approved):

PR #488, opened 2026-08-10
debezium/dbz#2395 Reject an OAuth2 private key that is not a data URI

PDGGK:
Closes debezium/dbz#2395.

## What

`OAuth2AuthHandler.validate` did not check that the OAuth2 private key is a `data:application/json;base64,` URI before slicing the prefix off it. `indexOf` returns `-1` when the prefix is absent, and `-1 + prefix.length()` is a valid offset, so a bare base64 key was **silently truncated by 28 characters** and then decoded — rather than being reported as the wrong shape.

The validator now requires the prefix at position 0.

## Scope, stated accurately

An earlier version of this description said an invalid key was simply accepted. @pxcamus tested it against a live Pulsar instance and showed that is not the whole story: such a connection *is* eventually marked invalid, because the Pulsar client rejects it while building its URI. The accurate framing is therefore narrower —

- the key passes **this** validator, whose job is to catch exactly this
- the failure surfaces later, from inside the Pulsar client, as a less obvious error
- for a bare base64 key the 28-character truncation happens first, so what gets reported is a decode/JSON failure rather than "this is not a data URI"

So this is a diagnosability fix, not a "bad credentials are accepted" fix.

## Also

Drops the dead `issuerUrl` assignment at the end of `validate()` — it was the last statement in the method and the value was never read. Included here at review request, since the branch needed a force-push anyway.

## Tests

`OAuth2AuthHandlerTest`, 8 cases. The ones that fail without the fix:

| case | why it matters |
|---|---|
| bare base64 whose first 28 chars are padding | the remainder decodes to perfectly valid credentials, so it has to be rejected on the **prefix**, not on what decodes |
| `"junk" + prefix + valid base64` | a contains-check accepts this; it must be rejected at position 0 |
| key shorter than the prefix | the original truncation case |

The other five pin existing behaviour — non-base64 payload, payload that is not JSON, credentials missing `client_id`/`client_secret`, missing config values, and the happy path — and pass before and after.

Run with JDK 22: `Tests run: 8, Failures: 0`.

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/destination/pulsar/OAuth2AuthHandler.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/destination/pulsar/OAuth2AuthHandler.java
@@ -78,17 +78,17 @@ public class OAuth2AuthHandler implements PulsarAuthHandler {
         String privateKey = config.get("oauth2PrivateKey").toString();
         byte[] privateKeyBytes = null;
 
-        // indexOf returns -1 when the prefix is absent, and -1 + prefix.length() is a valid offset,
-        // so without this check the key is silently truncated by 28 characters instead of rejected.
-        int prefixIndex = privateKey.indexOf(PRIVATE_KEY_PREFIX);
-        if (prefixIndex < 0) {
-            LOGGER.warn("OAuth2 private key is not a data URI; expected it to contain '{}'", PRIVATE_KEY_PREFIX);
+        // The key has to START with the prefix, not merely contain it: anything pasted in front of
+        // it would otherwise pass validation here and only fail later when the Pulsar client builds
+        // its URI. Without any check at all the key is silently truncated by 28 characters.
+        if (!privateKey.startsWith(PRIVATE_KEY_PREFIX)) {
+            LOGGER.warn("OAuth2 private key is not a data URI; expected it to start with '{}'", PRIVATE_KEY_PREFIX);
             throw new IllegalArgumentException(
                     "OAuth2 Private Key must be a data URI of the form '" + PRIVATE_KEY_PREFIX + "<base64-encoded JSON credentials>'");
         }
 
         try {
-            String extracted = privateKey.substring(prefixIndex + PRIVATE_KEY_PREFIX.length());
+            String extracted = privateKey.substring(PRIVATE_KEY_PREFIX.length());
             privateKeyBytes = Base64.getDecoder().decode(extracted);
         }
         catch (IllegalArgumentException e) {
@@ -110,6 +110,5 @@ public class OAuth2AuthHandler implements PulsarAuthHandler {
             throw new IllegalArgumentException("OAuth2 Private Key JSON must contain both 'client_id' and 'client_secret' fields");
         }
 
-        String issuerUrl = config.get("oauth2IssuerUrl").toString();
     }
 }
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/destination/pulsar/OAuth2AuthHandler.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/destination/pulsar/OAuth2AuthHandler.java
@@ -25,6 +25,9 @@ public class OAuth2AuthHandler implements PulsarAuthHandler {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(OAuth2AuthHandler.class);
 
+    /** Format as per the Pulsar doc -> data:application/json;base64,WaVf/RB7ZAW7pMG3rhxnL3zC2QFqyyLaKl5W6JwWdhw= */
+    private static final String PRIVATE_KEY_PREFIX = "data:application/json;base64,";
+
     /**
      * Configures the provided {@link PulsarAdminBuilder} with the authentication settings
      * required by this handler.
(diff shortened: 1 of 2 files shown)


mfvitale:
@pxcamus could I ask to take a look?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/connection/destination/pulsar/OAuth2AuthHandler.java
```diff
@@ -75,10 +78,17 @@ public void validate(Map<String, Object> config) throws IllegalArgumentException
         String privateKey = config.get("oauth2PrivateKey").toString();
         byte[] privateKeyBytes = null;
 
+        // indexOf returns -1 when the prefix is absent, and -1 + prefix.length() is a valid offset,
+        // so without this check the key is silently truncated by 28 characters instead of rejected.
+        int prefixIndex = privateKey.indexOf(PRIVATE_KEY_PREFIX);
+        if (prefixIndex < 0) {
```
pxcamus: @PDGGK Could you please do this:
```java
if (!privateKey.startsWith(PRIVATE_KEY_PREFIX)) { throw... }
```
Otherwise if the user mistakenly copies something before the key prefix, validation will pass here, but the actual Pulsar authentication will fail. Thanks!

pxcamus:
@PDGGK Thanks for catching that! A couple notes: I tested locally against my Pulsar instance, and an invalid key does pass validation before your fix, but is rejected by the Pulsar client when building its URI. So the UI does mark the connection invalid. But your fix is of course a valid improvement.

`issuerUrl` is indeed a dead assignment that could be clean up.

`openid`/`kerberos` are placeholders for schemes not yet implemented. So with a `@Named("OPENID")` bean implementing the `PulsarAuthHandler` interface, the factory will resolve it as-is. Thanks!

pxcamus (changes requested):

PDGGK:
@pxcamus thanks — you're right, and `startsWith` is strictly better. Pushed adbf622.

`indexOf` accepted anything pasted in front of the prefix, so validation passed here and the failure only surfaced later inside the Pulsar client. Added a test for exactly that case — `"junk" + PREFIX + base64(credentials)` — which fails against the previous version with *"Expected IllegalArgumentException to be thrown, but nothing was thrown"*. 8/8 green now.

Two follow-ups on your other points, both of which correct things I had wrong:

**On the severity of the original bug.** Thanks for actually testing it against a live Pulsar instance — that is a better check than mine. Your finding that the connection does get marked invalid downstream means my PR description overstated the impact: I framed it as an invalid key being accepted, when the accurate framing is that it is accepted *by this validator* and rejected later with a less obvious error. Happy to reword the description if you would like it on record.

**On `issuerUrl`.** Agreed it is a dead assignment. I have deliberately left it out of this PR to keep the diff to the one behaviour being fixed — say the word and I will either add it here or open a separate PR.

**On `openid`/`kerberos`.** Understood, thanks for the explanation — placeholders resolved by a `@Named` bean is a deliberate design, so there is nothing to fix there. I withdraw that part of my original note.

pxcamus:
@PDGGK  Since you will have to amend and force-push anyway to address the bot comments, I would suggest adding that `issuerUrl` clean up with it. And if you can reword the description, go ahead please. Thanks for the quick turn around!

Otherwise LGTM -- I will approve after your push.

FYI @mfvitale

PDGGK:
@pxcamus done — pushed 190a899 (force-push, as you suggested). All four points:

- **DCO** — both commits now signed off, matching the commit author
- **Commit prefix** — both carry `debezium/dbz#2395`
- **`issuerUrl`** — dead assignment dropped; it was the last statement in `validate()` and the value was never read
- **Description** — reworded

On the description: I narrowed the claim rather than just softening it. Your live-Pulsar test showed my original framing ("an invalid key is accepted") was wrong, so it now says what is actually true — the key passes *this* validator, and the failure surfaces later from inside the Pulsar client as a less obvious error. For a bare base64 key the 28-character truncation happens first, so what the user sees is a decode/JSON failure rather than "this is not a data URI". That makes it a diagnosability fix, which is a smaller claim than I made originally.

8/8 green on JDK 22.

pxcamus (approved):

PDGGK:
Thanks for the approval! One note on the red `Run tests`, since it is now the only thing between this and a merge — I do not believe it is this PR, and here is why rather than just an assertion.

Three integration tests fail, and both failure modes are about data being absent, not about validation:

```
CatalogResourceIT.testGetCatalogWithoutVersion
CatalogResourceIT.testGetCatalogWithoutVersionAndWithType
  JSON path schemaVersion doesn't match.  Expected: not null  Actual: null

MonitoringResourceIT.queryPanelReturnsDataFromPrometheus
  JSON path series doesn't match.  Expected: not an empty collection  Actual: <[]>
```

This PR touches exactly two files — `OAuth2AuthHandler.java` and its unit test — and neither `CatalogResourceIT` nor `MonitoringResourceIT` mentions OAuth2 or Pulsar at all (grep: 0 hits in each). There is no path from a Pulsar connection validator to a catalog `schemaVersion` or a Prometheus series.

One thing I want to be straight about rather than overstate: I could **not** compare against an earlier run of my own branch. On the previous commit the `Run tests` job was *skipped*, because the sign-off check failed first — so this is the first time tests have actually executed here, and I have no before/after on the same branch to point at. What I do have is the diff scope above, and that the Prometheus one in particular reads like a scrape-window timing flake.

I cannot re-run jobs myself. Happy to push an empty commit to re-trigger if that is the convention here — just say the word rather than me churning the branch after you have already approved it.

mfvitale:
@PDGGK Failures are not related. Applied, thanks!

PR #493, opened 2026-08-13
debezium/dbz#2101 Add pipeline deployment lifecycle with Ansible-based container management

### LINKED ISSUE #2101: [Host Deployment ~6] Implement Ansible-based pipeline deployment service
Implement the Conductor-side pipeline deployment logic using Ansible ad-hoc commands as the execution transport. This provides a working end-to-end deployment flow before the Host Agent is available — the same service logic is reused when the Agent is built later (the Ansible calls simply get swapped for HTTP calls).

### What

- **`HostPipelineController`** — implements `PipelineController` for host mode. Uses Ansible's `shell` module via `ProcessBuilder` to run Docker commands (`docker run`, `docker rm -f`, `docker stop`, `docker start`, `docker logs`) on the remote host. Config files are copied via the Ansible `copy` module. Runs on the existing dedicated thread pool.

- **`HostPipelineMapper`** — converts `PipelineFlat` into a Debezium Server `application.properties` string. Uses `TreeMap` + manual serialization for deterministic output and stable SHA-256 hashing (drift detection).

- **`HostDeploymentService`** — concurrency-safe host selection and port allocation. `PESSIMISTIC_WRITE` locks on all `READY` hosts sorted by ID (deadlock prevention), live `COUNT` query for load (no stale `pipeline_count` column), `MAX(serverPort)+1` for port allocation. `@Transactional(REQUIRES_NEW)` to release locks before the Ansible call. Pluggable via `DeployStrategy` interface with a default `RoundRobinStrategy`.

- **`HostDeploymentStatusPoller`** — `@Scheduled(every = "30s")` background poller. Checks container state via `ansible ... -m shell -a "docker inspect ..."`. Transitions: `DEPLOYING→RUNNING`, `RUNNING→FAILED`, `RUNNING→CONFIG_DRIFT`.

### Verification
- Full lifecycle test: deploy → running → stop → start → undeploy.
- Concurrency test: simultaneous deploys get different hosts and non-colliding ports.
- Config drift: tamper with the deployed config, verify poller sets `CONFIG_DRIFT`.

Tracks a sub-task of https://github.com/debezium/dbz/issues/2085
cc @mfvitale @kmos

d1vyanshu-kumar:
Fixes debezium/dbz#2101

## Description

Implements the full pipeline deployment lifecycle for host-mode environments using Ansible ad-hoc commands over SSH instead of the future Host Agent. This allows pipelines to be deployed, monitored, stopped, started, and undeployed on remote servers without blocking on the Host Agent implementation.

### Changes

**Pipeline Configuration Mapping**
- Add `HostPipelineMapper` to convert pipeline objects into flat `application.properties` format with source, sink, offset storage, schema history, transforms, and signal/notification defaults
- Apply platform-required defaults (signal channels, notification channels) after user config to prevent silent overrides, consistent with the operator `PipelineMapper`
- Map `debezium.format.key` and `debezium.format.value` from `JsonConverter` to `Json` to match Debezium Server's expected format enum

**Deployment Service & Host Selection**
- Add `HostDeploymentService` with transactional deployment creation, pessimistic-locked host selection, and port allocation
- Add `DeployStrategy` interface with `RoundRobinStrategy` (least-loaded, deterministic tie-breaking by host ID)
- Add `deployedAt` timestamp to `HostDeploymentEntity` for deployment age tracking
- Add `findReadyHosts()` query for container cleanup when deployment records are removed by cascade

**Pipeline Controller**
- Add `HostPipelineController` orchestrating deploy, stop, start, and undeploy operations via `AnsibleCommandRunner`
- Config files are copied to remote hosts using Ansible `copy` module; containers are managed using Ansible `shell` module running Docker commands
- Clean up existing deployment records and containers before redeployment to prevent UNIQUE constraint violations on `pipeline_id`
- Handle undeploy when the deployment DB record was already removed by `ON DELETE CASCADE` by falling back to `docker rm -f` on all ready hosts
- Add null guard on `Arc.container()` in CDI request context activation to support plain JUnit test execution

**Status Monitoring**
- Add `HostDeploymentStatusPoller` that checks container health every 30 seconds via `docker inspect`
- Implement a 5-minute grace period after deployment to allow Docker image pulls on fresh servers before marking containers as FAILED
- Detect config drift by comparing SHA-256 hash of deployed config against the live file on the remote host
- Reset `deployedAt` timestamp when transitioning back to DEPLOYING (restart scenario) to prevent false FAILED marking
- Add `ansible_become_timeout` (60s) to prevent Ansible `become` hangs on slow hosts

**Ansible Command Runner**
- Add `AnsibleCommandRunner` as a reusable wrapper for executing Ansible ad-hoc commands with structured result handling via sealed `CommandResult` interface

**Pipeline Deletion**
- Add `ON DELETE CASCADE` to `FK_host_deployment_pipeline` foreign key so deleting a pipeline from the UI automatically removes the associated `host_deployment` row without a constraint violation
- Add `@OnDelete(action = CASCADE)` on `HostDeploymentEntity.pipeline` to keep the JPA model consistent with the database schema

**Dev Profile Configuration**
- Override JDBC offset and schema history connection URLs under the `%dev` profile so that Debezium Server containers on remote hosts connect to the correct PostgreSQL instance instead of `localhost`

### Testing

- `HostPipelineMapperTest` — 17 unit tests covering source/sink/transform mapping, offset/schema history config, signal defaults ordering, format mapping, and SHA-256 hashing
- `HostPipelineControllerTest` — 17 unit tests for deploy/stop/start/undeploy orchestration, idempotent redeployment cleanup, and cascade undeploy fallback
- `HostDeploymentStatusPollerTest` — 12 unit tests covering all state transitions, grace period logic, config drift detection, and become timeout configuration
- `HostDeploymentServiceIT` — 15 integration tests against real PostgreSQL for deployment CRUD, status transitions, deployedAt lifecycle, pessimistic locking, ready host queries, and cascade deletion
- `AnsibleCommandRunnerTest` — unit tests for command construction and result parsing
- `RoundRobinStrategyTest` — unit tests for least-loaded selection and tie-breaking

## PR Checklist


d1vyanshu-kumar:
image
image
image

d1vyanshu-kumar:
Hi @mfvitale, @kmos, could you take a look at this when you get a chance?

One thing I wanted to mention, during our call we discussed using file-based offset/schema storage for Host mode, but after working through the implementation I went with the same JDBC-based approach (`JdbcOffsetBackingStore` / `JdbcSchemaHistory`) that we already use in Operator mode. The main reason is that with file-based storage, offsets live inside the container filesystem, so if a container gets recreated during redeploy or moved to another host, those offsets are lost and Debezium would have to re-read the WAL from scratch. With JDBC storage the containers stay fully stateless - the offset bookmark lives in a database table, so redeploys just resume from where they left off.

It also kept things simpler on the implementation side since we reuse the existing `PipelineConfigGroup` config paths without needing extra volume mount logic over SSH. The `%dev` profile points to a Postgres on the target host for local testing, and in production it resolves through the same `OFFSET_JDBC_URL` env var as Operator mode.

That said, if you'd prefer the file-based approach instead, I'm happy to rework it, just let me know!

mfvitale:
> That said, if you'd prefer the file-based approach instead, I'm happy to rework it, just let me know!

We should go with file based offset mounting as a volume on the target host.

kmos:
> > That said, if you'd prefer the file-based approach instead, I'm happy to rework it, just let me know!
> 
> We should go with file based offset mounting as a volume on the target host.

I agree. The target host shouldn't change like a k8s pod or volume. I think it's fine for now

debezium-platform-conductor/src/main/resources/db/migration/V3.6.0__create_heartbeat_table.sql
```diff
@@ -31,13 +31,15 @@ create table host_deployment (
...
 
 alter table if exists host_deployment
     add constraint FK_host_deployment_pipeline
     foreign key (pipeline_id)
-    references pipeline;
+    references pipeline
+    on delete cascade;
 
 alter table if exists host_deployment
     add constraint FK_host_deployment_host_status
```
mfvitale, line 43: You need to create another change file. You cannot modify an old one.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/strategy/DeployStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+     * Selects a host from the given list of locked, ready hosts.
+     *
+     * @param readyHosts  all hosts with {@code READY} status, sorted by ID,
+     *                    locked with {@code PESSIMISTIC_WRITE}
+     * @param loadCounter function that returns the number of active
+     *                    deployments for a given host ID
+     * @return the selected host entity
+     * @throws io.debezium.DebeziumException if no hosts are available
+     */
+    HostStatusEntity select(List<HostStatusEntity> readyHosts, LoadCounter loadCounter);
```
mfvitale: `LoadCounter` is specific for the round-robin load based implementation. Should not be in the interface

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineMapper.java
```diff
@@ -0,0 +1,381 @@
...
+ * but produces a flat key=value string instead of a Kubernetes CRD.
+ *
+ * <p><strong>Deterministic output:</strong> A {@link TreeMap} ensures
+ * alphabetical key ordering, and manual serialization (not
+ * {@link java.util.Properties#store}) avoids the non-deterministic
+ * timestamp comment. Both are critical for stable SHA-256 hashing
+ * used in config drift detection.
+ */
+@ApplicationScoped
+public class HostPipelineMapper {
```
mfvitale, line 36: I think that you can simply this a lot using the `io.debezium.platform.environment.operator.PipelineMapper.map(pipelineFlat)` and the call the `asConfiguration` on the `DebeziumServer` returned from the `map` method. 

This is effectively what the [operator](https://github.com/debezium/debezium-operator/blob/d5d53bc40ef72a722c8af8ac4976fb5c1099c5c1/debezium-operator-core/src/main/java/io/debezium/operator/core/dependent/ConfigMapDependent.java#L29) does to create the config map with the `application.properties`

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
```diff
@@ -0,0 +1,372 @@
...
+                deploymentService.deleteDeployment(existing.getId());
+            });
+
+            HostPipelineMapper.MappedConfig mappedConfig = pipelineMapper.map(pipeline);
+
+            HostDeploymentService.HostAllocation allocation = deploymentService.allocateHostAndPort();
+            String sshAlias = allocation.hostStatus().getSshAlias();
+            int port = allocation.allocatedPort();
+
+            String containerName = hostConfig.containerNamePrefix() + pipelineId;
```
mfvitale: We should use the pipeline the name ad we do for the k8s

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
```diff
@@ -0,0 +1,372 @@
...
+            String configDir = hostConfig.configBasePath() + PATH_SEPARATOR + pipelineId;
+            String configPath = configDir + PATH_SEPARATOR + CONFIG_FILE_NAME;
+
+            deploymentService.createDeployment(
+                    pipelineId, allocation.hostStatus().getId(),
+                    containerName, hostConfig.debeziumServerImage(),
+                    port, mappedConfig.configHash());
+
+            CommandResult mkdirResult = ansibleRunner.createDirectory(sshAlias, configDir);
```
mfvitale: I suppose that all commands below this point are Ansible related. Since we will replace this part with the host agent, I could see an interface with the `start`, `stop`, `restart` and in this PR the implementation is done via Ansible and then the other one that will call the API?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
```diff
@@ -0,0 +1,372 @@
...
+ * </ul>
+ *
+ * <p>Long-running Ansible calls are submitted to a dedicated thread pool
+ * to avoid blocking the reactive event thread.
+ *
+ * <p>When the Host Agent is built (sub-issue 7+), the Ansible calls in
+ * this class will be swapped for HTTP REST calls — all other logic
+ * (host selection, port allocation, status tracking) stays unchanged.
+ */
+@Dependent
```
Naros: Should this be `@ApplicationScoped`?

In CDI, for a `@Dependent` scoped bean, the container will create a new instance specifically for each observer method invocation, then destroy it afterward. The instance injected into `HostEnvironmentController`, the one holding the executor, is never notified when the shutdown event is observed.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
```diff
@@ -0,0 +1,208 @@
...
+     * Runs an Ansible ad-hoc command using the {@code copy} module
+     * to upload content to a file on the remote host.
+     *
+     * @param sshAlias the target host SSH alias
+     * @param content  the file content to upload
+     * @param destPath the absolute path on the remote host
+     * @return the command result
+     */
+    public CommandResult copyContent(String sshAlias, String content, String destPath) {
+        String copyArgs = "content='" + content + "' dest=" + destPath + " mode=0644";
```
Naros: The `content` argument is being wrapped here in single quotes; however, if any user-provided config value, such as a password or other connector property, contains a single quote, the Ansible copy module argument parsing will break.

We should either add escaping logic to guard against this, use Ansible's stdin, or write to a temp file and use `src=` rather than `content=`.

debezium-platform-conductor/src/main/resources/application.yml
```diff
@@ -208,6 +208,14 @@ quarkus:
         type: io.debezium.storage.jdbc.offset.JdbcOffsetBackingStore
         config:
           jdbc:
+            # Host-mode: JdbcOffsetBackingStore reads 'jdbc.connection.url' internally.
+            # Must point to a Postgres reachable from the remote Docker container.
+            # Override via HOST_OFFSET_JDBC_URL env var when the VM IP changes.
+            connection:
+              url: ${HOST_OFFSET_JDBC_URL:jdbc:postgresql://192.168.252.23:5432/testdb?loggerLevel=OFF}
```
Naros: It looks like the IP here might be specific to your environment? Same for `HOST_SCHEMA_HISTORY_JDBC_URL` below.

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/HostPipelineControllerTest.java
```diff
@@ -0,0 +1,562 @@
...
+    void undeploySkipsGracefullyWhenNoDeploymentExists() throws Exception {
+        when(deploymentService.findByPipelineId(999L)).thenReturn(Optional.empty());
+
+        CountDownLatch latch = new CountDownLatch(1);
+        // Use a short-lived approach: we need to verify the method completes
+        // Since there's nothing to await on the mock, we add a small delay
+        controller.undeploy(999L);
+
+        // Give the executor time to process
+        Thread.sleep(500);
```
Naros: Could we not use this instead?
```
assertThat(latch.await(ASYNC_WAIT_SECONDS, TimeUnit.SECONDS)).isTrue();
```

debezium-platform-conductor/src/main/resources/application.yml
mfvitale: @Naros  This part must be removed since we want to use file based offset on the host as default offset/schema storage

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
d1vyanshu-kumar: I've switched from content='...' to writing a local temp file and using src= instead. The config content never appears in the Ansible command string now, so single quotes (or any other characters) in user-provided values can't break argument parsing.

debezium-platform-conductor/src/main/resources/application.yml
d1vyanshu-kumar: @mfvitale, @kmos Agreed; switched to file-based offset and schema history storage. The host's data directory (/opt/debezium/data/<pipeline-name>) is bind-mounted into the container at /debezium/data. Offset data survives container restarts. Also removed the hardcoded 192.168.252.23 JDBC URLs from application.yml.

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/HostPipelineControllerTest.java
d1vyanshu-kumar: Thanks! Replaced all Thread.sleep(500) calls with CountDownLatch.await(ASYNC_WAIT_SECONDS, TimeUnit.SECONDS) , same pattern used in HostProvisioningServiceTest. No more timing-dependent flakiness.

d1vyanshu-kumar:
Hey @mfvitale @Naros @kmos  - pushed a commit addressing all the review feedback here is summary:

1. **Migration immutability** - Reverted V3.6.0 to its original heartbeat-only content and moved all host deployment DDL into a new V3.7.0.2 migration file. Flyway checksum stays intact for existing installations.

2. **LoadCounter out of DeployStrategy** - Removed it from the interface. The deployment count query now lives inside RoundRobinStrategy via direct EntityManager injection, so future strategies don't need to deal with load counting at all.

3. **HostPipelineMapper delegation** - Refactored to delegate to
PipelineMapper.map().asConfiguration().getAsMapSimple() and only override offset/schema storage from JDBC to file-based for host mode. 
   
4. **Container naming** - Now uses pipeline.getName() instead of a numeric ID prefix, matching the K8s convention. docker ps on the host shows the actual pipeline name.

5. **HostContainerRuntime interface** - Extracted deploy/undeploy/stop/start/logs into an interface. AnsibleContainerRuntime implements it. The controller depends only on the interface, so swapping to the Host Agent later won't require controller changes.

6. **@ Dependent → @ApplicationScoped** - Fixed the CDI scope on HostPipelineController so the shutdown observer fires on the real singleton that holds the running thread pool.

7. **copyContent injection fix** - Switched from content='...' to writing a local temp file and using src= instead. Config content never appears in the Ansible command string now.

8. **File-based offset/schema storage** - Host data directory is bind-mounted at /debezium/data. Removed the hardcoded JDBC URLs from application.yml. Verified offsets.dat is created on the host after the first snapshot.

9. **Thread.sleep → CountDownLatch**  Replaced all Thread.sleep(500) in async tests with CountDownLatch.await() for deterministic test behavior.

Also fixed a container permission issue that came up during live testing - the Debezium Server container couldn't write to the root-owned data directory. Per @mfvitale's suggestion, added --user $(id -u):$(id -g) to the docker run command so the container process matches the host directory owner. Tested and verified that offsets.dat is now created successfully.

As we discussed, I've opened a separate issue for the shared group provisioning improvement as a follow-up: [dbz#2451](https://github.com/debezium/dbz/issues/2451)

I have verified everything end-to-end on a live host server, and it is working smoothly. Please feel free to pull the latest changes and test it on your end, and let me know if anything else needs attention!

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+        this.hostConfig = hostConfig;
+    }
+
+    /**
+     * Result of host selection and port allocation.
+     *
+     * @param hostStatus    the locked and selected host entity
+     * @param allocatedPort the unique port assigned for this deployment
+     */
+    public record HostAllocation(HostStatusEntity hostStatus, int allocatedPort) {
```
kmos: move this record outside this class

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+        return deployment;
+    }
+
+    /**
+     * Finds the active deployment for a given pipeline.
+     */
+    @Transactional(REQUIRES_NEW)
+    public Optional<HostDeploymentEntity> findByPipelineId(Long pipelineId) {
+        return em.createQuery(
+                "SELECT d FROM host_deployment d WHERE d.pipeline.id = :pipelineId",
```
kmos: extract the query

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    /**
+     * Returns all deployments with a given status.
+     */
+    @Transactional(REQUIRES_NEW)
+    public List<HostDeploymentEntity> findByStatus(DeploymentStatus status) {
+        return em.createQuery(
+                "SELECT d FROM host_deployment d WHERE d.deploymentStatus = :status",
```
kmos: extract the query

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+     * @param pipelineId    the pipeline being deployed
+     * @param hostStatusId  the selected host
+     * @param containerName the Docker container name
+     * @param imageVersion  the Debezium Server image version
+     * @param serverPort    the allocated port
+     * @param configHash    SHA-256 hash of the deployed config
+     * @return the persisted deployment entity
+     */
+    @Transactional(REQUIRES_NEW)
+    public HostDeploymentEntity createDeployment(Long pipelineId, Long hostStatusId,
```
kmos: is it really necessary to return an `HostDeploymentEntity`? it could be void?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+                                                 String containerName, String imageVersion,
```
kmos: reduce the number of parameter creating a new record class. Try to follow the rule max 3 params

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+                pipelineId, hostStatusId, serverPort);
+
+        return deployment;
+    }
+
+    /**
+     * Finds the active deployment for a given pipeline.
+     */
+    @Transactional(REQUIRES_NEW)
+    public Optional<HostDeploymentEntity> findByPipelineId(Long pipelineId) {
```
kmos: Are really necessary all the information of the `HostDeploymentEntity` ? You may create a class `deployment`?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+                HostDeploymentEntity.class)
+                .setParameter("statuses", List.of(statuses))
+                .getResultList();
+    }
+
+    /**
+     * Returns all READY hosts without pessimistic locking.
+     */
+    @Transactional(REQUIRES_NEW)
+    public List<HostStatusEntity> findReadyHosts() {
```
kmos: same as this comment https://github.com/debezium/debezium-platform/pull/493/changes#r3805479903

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,250 @@
...
+                HostStatusEntity.class)
+                .setParameter("status", ProvisioningStatus.READY)
+                .getResultList();
+    }
+
+    /**
+     * Locks all READY hosts sorted by ID. Sorting prevents ABBA deadlocks
+     * when concurrent transactions lock the same set of rows.
+     */
+    private List<HostStatusEntity> lockAllReadyHosts() {
```
kmos, line 239: same https://github.com/debezium/debezium-platform/pull/493/changes#r3805479903

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
```diff
@@ -0,0 +1,232 @@
...
+    public AnsibleCommandRunner(Logger logger, HostConfigGroup hostConfig) {
+        this.logger = logger;
+        this.hostConfig = hostConfig;
+    }
+
+    /**
+     * Sealed result type for ad-hoc command execution.
+     */
+    public sealed interface CommandResult {
+        record Success(String output) implements CommandResult {
```
kmos: could you move it outside?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
```diff
@@ -0,0 +1,232 @@
...
+        java.nio.file.Path tempFile = null;
+        try {
+            tempFile = java.nio.file.Files.createTempFile("debezium-config-", ".properties");
+            java.nio.file.Files.writeString(tempFile, content, StandardCharsets.UTF_8);
+
+            String copyArgs = "src=" + tempFile.toAbsolutePath() + " dest=" + destPath + " mode=0644";
+            List<String> command = buildAdHocCommand(sshAlias, "copy", copyArgs);
+            return executeCommand(command);
+        }
+        catch (IOException e) {
```
kmos: If you want to catch the exception, please make sure not to lose the stack trace.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
```diff
@@ -0,0 +1,232 @@
...
+            logger.errorv(e, "Failed to create temp file for Ansible copy");
+            return new CommandResult.Failure("Failed to create temp file: " + e.getMessage());
+        }
+        finally {
+            if (tempFile != null) {
+                try {
+                    java.nio.file.Files.deleteIfExists(tempFile);
+                }
+                catch (IOException e) {
+                    logger.warnv("Failed to delete temp file {0}: {1}", tempFile, e.getMessage());
```
kmos: same as above

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
```diff
@@ -0,0 +1,232 @@
...
+            logger.debugv("Executing Ansible ad-hoc command: {0}", command);
+
+            process = launchProcess(command);
+
+            boolean finished = process.waitFor(AD_HOC_TIMEOUT_MINUTES, TimeUnit.MINUTES);
+            String output = captureOutput(process);
+
+            if (!finished) {
+                process.destroyForcibly();
+                String timeoutMessage = "Ansible ad-hoc command timed out after "
```
kmos: inline

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
```diff
@@ -0,0 +1,232 @@
...
+                logger.debugv("Ansible ad-hoc command succeeded");
+                return new CommandResult.Success(output);
+            }
+
+            logger.warnv("Ansible ad-hoc command failed with exit code {0}: {1}",
+                    exitCode, output);
+            return new CommandResult.Failure(output);
+        }
+        catch (IOException e) {
+            logger.errorv(e, "Failed to start Ansible ad-hoc process");
```
kmos: same as the previous comment related to stacktrace

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/strategy/RoundRobinStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+public class RoundRobinStrategy implements DeployStrategy {
+
+    private final EntityManager em;
+
+    public RoundRobinStrategy(EntityManager em) {
+        this.em = em;
+    }
+
+    @Override
+    public HostStatusEntity select(List<HostStatusEntity> readyHosts) {
```
kmos: Why using the entity? You could use a domain class

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/provisioning/AnsibleCommandRunner.java
```diff
@@ -0,0 +1,232 @@
...
+ * </ul>
+ *
+ * <p>Output is captured synchronously (ad-hoc commands are fast, unlike
+ * playbooks) with a configurable timeout. The merged stdout+stderr
+ * approach prevents pipe-buffer deadlocks.
+ *
+ * @see AnsibleHostProvisioner
+ */
+@ApplicationScoped
+public class AnsibleCommandRunner {
```
kmos, line 48: you could refactor using command pattern: https://refactoring.guru/design-patterns/command

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AnsibleContainerRuntime.java
```diff
@@ -0,0 +1,125 @@
...
+        this.hostConfig = hostConfig;
+    }
+
+    @Override
+    public void deploy(String host, String containerName, int port, String configContent, String image) {
+        String configDir = hostConfig.configBasePath() + PATH_SEPARATOR + containerName;
+        String configPath = configDir + PATH_SEPARATOR + CONFIG_FILE_NAME;
+        String dataDir = hostConfig.dataBasePath() + PATH_SEPARATOR + containerName;
+
+        // 1. Create config directory
```
kmos: Instead of using comment in the code, express your intentions: extract classes that describe the business

kmos (commented):
I have done a round of review, my suggestion is to take a look to the [following article](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) and internal links and apply a refactoring.

debezium-platform-conductor/src/main/resources/db/migration/V3.6.0__create_heartbeat_table.sql
```diff
@@ -2,48 +2,3 @@ CREATE TABLE IF NOT EXISTS heartbeat (
     id INTEGER PRIMARY KEY,
     timestamp TIMESTAMP NOT NULL DEFAULT now()
 );
-
--- Host-based pipeline deployment tables
--- Purely additive: no changes to existing tables, sequences, indexes, or constraints.
-
-create sequence host_status_SEQ start with 1 increment by 50;
-
-create sequence host_deployment_SEQ start with 1 increment by 50;
```
mfvitale, line 11: @d1vyanshu-kumar Existing file should not be modified. If you need to add column to an already existing table you just need to crate the alter DDL statement on a new migration file.

d1vyanshu-kumar: @mfvitale ok, I see it now. I moved the host tables out to a new file, but deleting them from V3.6.0 still counts as modifying it, so that's why this popped up again. My bad, makes sense now. 

I've reverted V3.6.0 and V3.7.0.2 back to their original state and added the deployed_at column + the ON DELETE CASCADE change through a new migration file instead, so neither of the existing files get touched. Let me know if that looks right. 🙂

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
d1vyanshu-kumar: Changed to void. The only real caller (`HostPipelineController.executeDeploy)` wasn't using the return value, it was being silently discarded. The IT tests that verified field values now query back via `requireByPipelineId()` instead.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
d1vyanshu-kumar: Created a lightweight Deployment domain record in [io.debezium.platform.domain](https://github.com/debezium/debezium-platform/blob/79a9ea8daa287af66d59b0eef50ca7c18789386e/debezium-platform-conductor/src/main/java/io/debezium/platform/domain/Deployment.java#L23). All public query methods now return Deployment records mapped via` Deployment.from()` at the service boundary. Controllers and pollers never touch JPA entities

d1vyanshu-kumar:
Hi @kmos, thanks for the detailed feedback and the architecture resources you shared. Reading through Uncle Bob's Clean Architecture article and the Command pattern doc really helped me see what you meant about keeping JPA entities from leaking across boundaries. Here's a summary of what I changed:

1. Introduced Host, Deployment, DeploymentRequest, and HostAllocation as domain records so controllers and pollers never touch JPA entities directly
2. Applied the Command pattern to AnsibleCommandRunner with ShellCommand, CopyCommand, and FileCommand, the runner now has a single execute() path
3. Replaced the numbered comments in AnsibleContainerRuntime.deploy() with self-documenting private methods
4. Extracted all JPQL queries to constants, changed createDeployment() to void, and preserved stack traces in all catch blocks

All unit tests and integration tests pass. Let me know if this looks good or if there's anything you'd like me to adjust.

debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/HostDeploymentEntity.java
```diff
@@ -34,6 +39,7 @@ public class HostDeploymentEntity {
     private Long id;
 
     @ManyToOne
```
mfvitale: Should this be `@OneToOne` since it is `unique`?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -0,0 +1,248 @@
...
+ * deployments exist. Uses a live query (not a stale column) to prevent
+ * port collision under concurrent writes.
+ *
+ * <p><strong>Domain boundary:</strong> All public methods return domain
+ * classes ({@link Deployment}, {@link Host}, {@link HostAllocation})
+ * instead of JPA entities. This prevents persistence-layer details from
+ * leaking into controllers and pollers.
+ */
+@ApplicationScoped
+public class HostDeploymentService {
```
mfvitale: Could you please extends the `AbstractService` similar to `io.debezium.platform.domain.HostStatusService`?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostAllocation.java
```diff
@@ -0,0 +1,19 @@
...
+ * Result of host selection and port allocation.
+ *
+ * <p>Returned by {@link HostDeploymentService#allocateHostAndPort()}.
+ * Uses the {@link Host} domain class instead of the JPA entity so that
+ * callers never handle persistence-layer types.
+ *
+ * @param host          the selected host
+ * @param allocatedPort the unique port assigned for this deployment
+ */
+public record HostAllocation(Host host, int allocatedPort) {
```
mfvitale: Let's be consistent and use the `views` from blaze persistence

kmos:
> Hi @kmos, thanks for the detailed feedback and the architecture resources you shared. Reading through Uncle Bob's Clean Architecture article and the Command pattern doc really helped me see what you meant about keeping JPA entities from leaking across boundaries. Here's a summary of what I changed:
> 
> 1. Introduced Host, Deployment, DeploymentRequest, and HostAllocation as domain records so controllers and pollers never touch JPA entities directly
> 2. Applied the Command pattern to AnsibleCommandRunner with ShellCommand, CopyCommand, and FileCommand, the runner now has a single execute() path
> 3. Replaced the numbered comments in AnsibleContainerRuntime.deploy() with self-documenting private methods
> 4. Extracted all JPQL queries to constants, changed createDeployment() to void, and preserved stack traces in all catch blocks
> 
> All unit tests and integration tests pass. Let me know if this looks good or if there's anything you'd like me to adjust.

:rocket: let's see what it's the final result ;)

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
d1vyanshu-kumar: Done!

debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/HostDeploymentEntity.java
d1vyanshu-kumar: Changed it, thank you!

d1vyanshu-kumar:
> 🚀 let's see what it's the final result ;)

Hi @kmos  here we go:


Screenshot 2026-08-26 at 12 47 34 AM

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
```diff
@@ -0,0 +1,301 @@
...
+
+            String containerName = pipeline.getName();
+
+            deploymentService.createDeployment(
+                    pipelineId, allocation.host().getId(),
+                    new DeploymentRequest(containerName, hostConfig.debeziumServerImage(),
+                            port, mappedConfig.configHash()));
+
+            // Delegate all infrastructure work to the container runtime
+            containerRuntime.deploy(sshAlias, containerName, port,
```
mfvitale: Give the `sshAlias` and `port` are in the `HostAllocation` record could we directly pass it?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
```diff
@@ -0,0 +1,301 @@
...
+        else {
+            task.run();
+        }
+    }
+
+    /**
+     * {@link LogReader} implementation that reads Docker container logs
+     * via the {@link HostContainerRuntime}.
+     */
+    private class HostDockerLogReader implements LogReader {
```
mfvitale: Could you move this out?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
d1vyanshu-kumar: Good point! Removed the redundant `sshAlias` and `port` local variables.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
d1vyanshu-kumar: Done! Extracted `HostDockerLogReader` into its own standalone package-private class file.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
mfvitale: Maybe I was not completely clear. I meant to directly pass the `Allocation` object to the deploy method instead of passing both `port` and `sshAlias`

d1vyanshu-kumar: My bad!  Updated `HostContainerRuntime.deploy` to accept HostAllocation directly now. Could you please take another look?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostDeploymentStatusPoller.java
```diff
@@ -0,0 +1,251 @@
...
+                }
+                logger.warnv("Container {0} on {1} is not running after deploy (grace period elapsed), marking FAILED",
+                        containerName, sshAlias);
+                deploymentService.updateStatus(deploymentId, DeploymentStatus.FAILED);
+                return;
+            }
+
+            if (!containerRunning && currentStatus == DeploymentStatus.RUNNING) {
+                int failures = consecutiveFailures.merge(deploymentId, 1, Integer::sum);
+                if (failures >= FAILURE_THRESHOLD) {
```
mfvitale: Could you check  if you can use the Retryable like in https://github.com/debezium/debezium-platform/blob/main/debezium-platform-conductor%2Fsrc%2Fmain%2Fjava%2Fio%2Fdebezium%2Fplatform%2Fenvironment%2Fwatcher%2Fconsumers%2FOutboxParentEventConsumer.java#L69

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
```diff
@@ -0,0 +1,246 @@
...
+            String sshAlias = deployment.getSshAlias();
+            String containerName = deployment.getContainerName();
+            containerRuntime.undeploy(sshAlias, containerName);
+
+            // Hard-delete the deployment record (frees UNIQUE constraint + port)
+            deploymentService.deleteDeployment(deployment.getId());
+            logger.infov("Pipeline {0} undeployed from host {1}", pipelineId, sshAlias);
+        }
+        else {
+            // Deployment DB record was already removed by ON DELETE CASCADE when pipeline row was deleted.
```
mfvitale: So effectively we should remove the cascade on the DB so we find the deployment and then we correctly stop the container first and then run the undeploy.

mfvitale (changes requested):

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostDeploymentStatusPoller.java
d1vyanshu-kumar: Done! PTAL.

debezium-platform-conductor/src/main/resources/db/migration/V3.7.0.2__add_deployed_at_cascade_delete.sql
```diff
@@ -0,0 +1,12 @@
...
+
+alter table if exists host_deployment
+    add column if not exists deployed_at timestamp(6) with time zone not null default now();
+
+alter table if exists host_deployment
+    drop constraint if exists FK_host_deployment_pipeline;
+
+alter table if exists host_deployment
+    add constraint FK_host_deployment_pipeline
+    foreign key (pipeline_id)
```
mfvitale: Given that we removed the `On CASCADE` I think we should remove this constraint otherwise we get a foreign key violation. Isn't it?

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherTestProfile.java
```diff
@@ -52,6 +52,7 @@ public Map<String, String> getConfigOverrides() {
         return Map.of(
                 "platform.deployment.mode", "host",
                 "platform.host.ssh-config-path", TEMP_SSH_CONFIG.toString(),
+                "conductor.descriptors.volume-source", "true",
```
mfvitale: Why this?

d1vyanshu-kumar: @mfvitale, Actually, the %test profile in `application.yml `defaults `conductor.descriptors.volume-source to `false`, which triggers Quarkus to start an ORAS DevServices container (Zot registry) during test bootstrap.

Setting "conductor.descriptors.volume-source", "true" in `SshConfigWatcherTestProfile` (matching HostModeTestProfile) disables the ORAS download mode so the SSH config watcher tests don't spin up an unnecessary container registry, keeping the test execution fast and lightweight. If you want, then I will revert this.

mfvitale:
@d1vyanshu-kumar there is already `V3.7.0.2__add_alerting.sql` so we should aim for `V3.7.0.3__add_deployed_at.sql`

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/HostDeploymentServiceIT.java
```diff
@@ -348,37 +349,51 @@ public void shouldFindReadyHosts() {
...
                 seededPipeline3Id, seededHostId,
                 new DeploymentRequest(
                         "debezium-pipeline-" + seededPipeline3Id,
                         "quay.io/debezium/server:latest",
                         9099, "cascadehash"));
 
-        // Query back to get the deployment ID
+        // Attempting to delete the pipeline directly must fail:
+        // the FK constraint (without CASCADE) blocks it while a deployment still references it.
+        PipelineEntity pipeline = em.find(PipelineEntity.class, seededPipeline3Id);
```
mfvitale: This is not a valid test. Without the FK the pipeline must be deleted without problems.

mfvitale (changes requested):

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/HostDeploymentServiceIT.java
d1vyanshu-kumar: You're right, removed the FK violation test and simplified to a single `shouldDeleteDeploymentRecord`() that only verifies `deleteDeployment`() works. Also renamed the migration to V3.7.0.3 and added the FK drop. PTAL.

mfvitale (approved):
@d1vyanshu-kumar Great work! Applied, thanks!

PR #504, opened 2026-08-18
Dbz#2351 Support other Pipeline status rather then FAILED

### LINKED ISSUE #2351: Support other Pipeline status rather then `FAILED`
Currently we set the pipeline status only on error case, we need to set also the other states.

mfvitale:
Fixes debezium/dbz#2351

## Description
This pull request introduces a set of changes to improve the health monitoring and status reconciliation of Debezium pipelines, particularly in Kubernetes operator deployments. The key enhancements include the addition of configurable health probe settings, a robust mechanism for mapping Kubernetes CR conditions to pipeline statuses, and improved status event handling across the environment and domain layers.

## PR Checklist


debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineStatusWatcher.java
```diff
@@ -0,0 +1,231 @@
...
+        // The operator never sets Running=False; when a pod crashes (CrashLoopBackOff),
+        // Ready simply flips back to False. Detect this by checking the transition:
+        // previously RUNNING + now Ready=False (without Running condition) = FAILED.
+        var failedViaTransition = false;
+        if (resolvedStatus.isEmpty()
+                && oldStatus.filter(s -> s == PipelineStatus.RUNNING).isPresent()
+                && hasCondition(newDs, CONDITION_READY, Condition.FALSE)) {
+            resolvedStatus = Optional.of(PipelineStatus.FAILED);
+            failedViaTransition = true;
+        }
```
Naros, line 140: I think this heuristic will report `FAILED` for ordinary redeploys, not just crashes?

`DebeziumServerReconciler.initializeNotReadyStatus()` replaces the entire condition list with a single `ServerNotReadyCondition` (`Ready=False`, message `"Server %s deployment in progress"`) on *every* reconcile where the dependent workflow isn't fully ready; not only on a crash. So a plain pipeline edit goes:

1. `PipelineConsumer.accept` to  `updateStatus(id, DEPLOYING, null)` to  `pipelines.deploy(p)`
2. operator reconciles; conditions become `[Ready=False]`
3. informer `onUpdate(old=[Ready=True, Running=True], new=[Ready=False])` to `mapConditionsToStatus(old)` = `RUNNING`, `mapConditionsToStatus(new)` = empty, `hasCondition(new, Ready, False)` = true to  `FAILED` with `"Pipeline failed unexpectedly. Check the pipeline logs for details."`
4. …then `RUNNING` again once the deployment settles.

That clobbers the `DEPLOYING` status this PR introduces and persists a misleading error message on every edit.

The test `reconcileOnUpdateResolvesFailedOnCrashLoopBackOff` looks like it encodes this same case rather than a crash; the CR it asserts `FAILED` for carries the message `"Server database-migration is being deployed"`.

Conditions alone can't separate a crash from a redeploy, but `metadata.generation` can: a bumped generation means the spec changed (redeploy in progress), while `Ready` flipping to `False` with an unchanged generation is a genuine runtime failure. Suppressing the transition-`FAILED` while the persisted status is `DEPLOYING` would also work. WDYT?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineStatusWatcher.java
```diff
@@ -0,0 +1,231 @@
...
+                        @Override
+                        public void onDelete(DebeziumServer obj, boolean deletedFinalStateUnknown) {
+                            // no-op: deletion is handled by the undeploy flow
+                        }
+                    });
+            LOGGER.info("DebeziumServer CR status watcher started successfully");
+        }
+        catch (Exception e) {
+            LOGGER.warn("Unable to start DebeziumServer CR status watcher: {}", e.getMessage());
+        }
```
Naros, line 114: Should this log the stack trace as well, or even be an error?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineStatusWatcher.java
```diff
@@ -0,0 +1,231 @@
...
+                .collect(Collectors.toMap(Condition::getType, c -> c));
+
+        var ready = conditionMap.get(CONDITION_READY);
+        if (ready != null && Condition.TRUE.equals(ready.getStatus())) {
+            var running = conditionMap.get(CONDITION_RUNNING);
+            if (running != null && Condition.FALSE.equals(running.getStatus())) {
+                return Optional.of(PipelineStatus.STOPPED);
+            }
+            return Optional.of(PipelineStatus.RUNNING);
+        }
```
Naros, line 177: Question: Reading the operator, I'm not sure `Ready=True` + `Running=False` is actually reachable.

`DeploymentDependent.desiredReplicas()` returns `0` for a stopped server, and `DeploymentReady.isMet()` is:

```java
Objects.equals(deployment.getSpec().getReplicas(), deployment.getStatus().getReadyReplicas())
```

For a Deployment scaled to zero, `status.readyReplicas` is omitted (`null`), so `Objects.equals(0, null)` is `false`, the workflow never reports ready, the reconciler takes the `initializeNotReadyStatus` branch and emits only `Ready=False`. `ServerReadyCondition` + `ServerStoppedCondition` would then never be written.

If that's right, stopping a pipeline goes `RUNNING => FAILED` via the transition rule in my other comment rather than `RUNNING => STOPPED`, which is the main scenario in debezium/dbz#2351. The unit tests here build the condition lists by hand, so they wouldn't catch it.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineStatusWatcher.java
mfvitale: > I think this heuristic will report FAILED for ordinary redeploys, not just crashes?

Yes, this is a good catch. I noticed that we lack some fine-grain statuses on the Operator and tried to extract something from the Platform. 

> but metadata.generation can 

I dint't thought to it! 

So effectively if a restart happens `[Ready=False]` and if the generation is changed this means that a `DEPLOYING` status due to a config change, otherwise we can treat it as a `FAILURE` right?

mfvitale: Initially I was thinking to add more status/messages on the operator side but then I decided to first look how the things works here and eventually move more detailed status there.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineStatusWatcher.java
mfvitale: Well, I tried the stopping feature on the `operator` and seems there are two bugs in it. 

- Reconcile never triggered when adding the `debezium.io/stop` annotation. 
This because the `ControllerConfiguration.generationAwareEventProcessing` defaults to `true` so a change to metadata [never increment the generation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource) and the reconciliation is never called. 
- Exactly the bug you reported about the `status.readyReplicas`

So I think that this should be solved on the operator side. WDYT?

In particular I think we should set to `0` when the replica is `null` and then let the reconcile work event when there is a metadata change. WDYT?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineStatusWatcher.java
mfvitale: @Naros I now used the `metadata.generation` to differentiate between a re-deploy from a crash.

Naros (approved):
This LGTM, feel free to merge if you're comfortable with it.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/OperatorPipelineStatusWatcher.java
mfvitale: Opened a bug https://github.com/debezium/dbz/issues/2528

mfvitale:
@indraraj Could you create the relative issue to also show the `DEPLOYING` and `RUNNING` status?

PR #507, opened 2026-08-24
debezium/dbz#2429 Avoid creating caches in core repo PRs

### LINKED ISSUE #2429: Adjust GHA to use reference-branch caches only
One of the issues with our current CI workload on the debezium/debezium repo is that we are constantly evicting caches from our 10GB budget because our Maven caches are ~2GB, and we generate a 2GB budget for each and every PR we build.

## Proposal 

Rather than have each PR create its own independent 2GB cache, we have the PR workflow always download the cache from its reference branch (main, 3.6, etc). We apply a marker strategy after the restore, then build our dependency pass.

The idea is that the reference branch cache is always restored only, with no saves.
Any artifacts downloaded after the restore will be bundled and added as a PR artifact.

Image

Subsequent jobs, like in our PR workflow, then overlay the PR artifact so that every 30+ jobs get the desired dependency cache without redownloads. We only pay for the cache per reference branch (2GB for main, 2GB for 3.6, etc.), while storing the delta per PR.

Artifacts would have a 1-day retention policy, so after 1 day, any change to the PR would require the dependency pass to redownload the artifact and create a new attachment on the PR.

Naros:
Fixes debezium/dbz#2429

## Description
Aligns with https://github.com/debezium/debezium/pull/7883, prevents creating cache in core repo.

## PR Checklist

### DIFF
--- a/.github/actions/build-cache/action.yml
+++ b/.github/actions/build-cache/action.yml
@@ -106,9 +106,14 @@ runs:
           echo "cache_key=maven-debezium-platform-conductor-${CORE_HASH}-${OPERATOR_HASH}-${DB2_HASH}-${INFORMIX_HASH}-${QUARKUS_HASH}" >> "$GITHUB_OUTPUT"
         fi
 
+    # Restore only.  The cache is written by the "Save Maven cache" step at the end of
+    # this action, and only when this action runs in this repository's own CI.  When it
+    # is driven by debezium/debezium (which supplies maven-cache-key) the dependencies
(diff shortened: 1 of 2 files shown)


mfvitale (approved):

mfvitale:
@indraraj are the FE failure expected?

indraraj:
Looks to be a network issue, I have triggered the rerun. lets see

indraraj:
Anyway, I don't see changes to the related

indraraj (approved):

PR #518, opened 2026-08-27
debezium/dbz#2509 Fix flaky SshConfigWatcherServiceIT with atomic file writes

### LINKED ISSUE #2509: [Platform] Flaky SshConfigWatcherServiceIT due to non-atomic file write
`SshConfigWatcherServiceIT.shouldMarkRemovedHostWhenDeletedFromFile` fails intermittently on loaded CI runs with:

```
Expected size: 2 but was: 3 in:
[HostStatus(id = 3), HostStatus(id = 1), HostStatus(id = 2)]
```

Reported in [debezium/debezium#7848](https://github.com/debezium/debezium/actions/runs/32216990521/job/95962869727?pr=7848).

### Root cause

The test writes the SSH config file using `Files.writeString()`, which truncates and rewrites the file in place. This is not atomic — there is a brief window where the file is empty or partially written on disk.

On a loaded CI machine, the background `WatchService` watcher thread can detect the file change and fire a debounced `reconcile()` call that reads the file during this truncate-then-write window. The parser then sees an empty (or partial) file, and reconciliation marks all hosts as `REMOVED`, corrupting the database state before the test assertion runs.

On a quiet local machine or isolated CI job, the race window is never hit because the watcher's debounced call lands after the write has fully completed.

d1vyanshu-kumar:
Fixes debezium/dbz#2509

## Description

`SshConfigWatcherServiceIT.shouldMarkRemovedHostWhenDeletedFromFile` fails intermittently on loaded CI runs ([debezium/debezium#7848](https://github.com/debezium/debezium/actions/runs/32216990521/job/95962869727?pr=7848)). The test modifies the SSH config file using `Files.writeString()`, which truncates the file before rewriting it. On a loaded machine, the background `WatchService` watcher can read the file during this brief truncate-then-write window and see an empty or partial file, leading to incorrect reconciliation state.

The fix makes file writes in the test atomic by writing to a temporary file first, then using `Files.move()` with `ATOMIC_MOVE`. This guarantees any concurrent reader always sees a complete file.

### Changes

- Extract `writeConfigAtomically()` helper in `SshConfigWatcherServiceIT`
- Replace all `Files.writeString()` calls with the atomic helper
- Add comment explaining the motivation (per review feedback)

## PR Checklist

### DIFF
--- a/debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherServiceIT.java
+++ b/debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/discovery/SshConfigWatcherServiceIT.java
@@ -9,6 +9,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.List;
 
 import jakarta.inject.Inject;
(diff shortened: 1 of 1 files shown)


mfvitale (approved):

mfvitale:
@d1vyanshu-kumar Applied, thanks!

PR #519, opened 2026-08-28
debezium/dbz#2271: Add validation for fully qualified table names in signal table data verification

### LINKED ISSUE #2271: NPE in conductor when signal data collection verification is called without a table name
## Bug report

  **What Debezium connector do you use and what version?**

  Debezium Platform (conductor), nightly build, 3.7.0-SNAPSHOT.

  ---

  **What behavior do you expect?**

  `POST /api/sources/signals/verify` with no fully-qualified table name in the request
  should be rejected with a clear validation error (e.g. HTTP 400 "table name is required"),
  not cause a server-side NullPointerException.

  ---

  **What behavior do you see?**

  When the table name is omitted, the conductor throws an NPE internally and the endpoint
  returns a response whose `message` is the meaningless string `"content"`:

  ```json
  {
      "exists": false,
      "message": "content"
  }
  ```

  Observed via the Debezium JBang CLI before `--table` was made mandatory
  (debezium/jbang-catalog#19):

  ```bash
  $ debezium source verify-signals --connection=2
  Signal collection table is not available — content
  ```

  With a table name supplied, the same endpoint responds correctly:

  ```bash
  $ debezium source verify-signals --connection=2 --table="public.signaling_table"
  Signal collection table is not available — Signal data collection not present or misconfigured
  ```

  The CLI now requires `--table` as a client-side workaround, but the conductor API itself
  still accepts the request and NPEs, so any direct API consumer can hit this.

  Discussion: https://debezium.zulipchat.com/#narrow/channel/392459-community-platform/topic/NPE.20when.20querying.20the.20conductor/with/611338069

pxcamus:
Fixes debezium/dbz#2271<the Debezium Issue Number>

## Description
This pull request hardens the signal data collection verification endpoint (JDBC-only) against requests that omit the fully qualified table name. POST /api/sources/signals/verify previously passed a null name through to TableId.parse, where a NullPointerException was caught and returned to the caller as HTTP 200 with the message "content" — a successful-looking response carrying a meaningless error. The key changes are rejection at the API boundary through bean validation, a defensive guard in the inspector so a parser failure can no longer be reported as a verification result, and test coverage at both layers.

## PR Checklist

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/api/SourceResource.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/api/SourceResource.java
@@ -105,6 +105,7 @@ public class SourceResource {
 
     @Operation(summary = "Verify that signal data collection is configured correctly")
     @APIResponse(responseCode = "200", content = @Content(mediaType = APPLICATION_JSON, schema = @Schema(implementation = SignalDataCollectionVerifyResponse.class, type = SchemaType.OBJECT)))
+    @APIResponse(responseCode = "400", description = "Invalid request, the connection id and the fully qualified table name are both required", content = @Content(mediaType = APPLICATION_JSON, schema = @Schema(type = SchemaType.OBJECT)))
     @POST
     @Path("/signals/verify")
     public Response verifySignalConfiguration(@NotNull @Valid SignalCollectionVerifyRequest request) {
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/SignalCollectionVerifyRequest.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/data/dto/SignalCollectionVerifyRequest.java
@@ -5,7 +5,10 @@
  */
 package io.debezium.platform.data.dto;
 
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+
 public record SignalCollectionVerifyRequest(
-        Long connectionId,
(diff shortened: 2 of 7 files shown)


pxcamus:
@mfvitale when you have time. I will need to open a follow up issue for @indraraj , since the Stage UI now needs to read the response. Which poses the question: is a boolean the right shape? But I think it is out of scope. WDYT?

mfvitale:
@pxcamus Thanks for the PR! 

> I will need to open a follow up issue for @indraraj , since the Stage UI now needs to read the response

Yes, sure. 

>Which poses the question: is a boolean the right shape? But I think it is out of scope. WDYT?

For this case I think that if the HTTP status is not `200` then the error is explained in the body. As for all error so I suppose we are good with the current return value. @indraraj Do you confirm?

mfvitale (approved):

indraraj:
> For this case I think that if the HTTP status is not `200` then the error is explained in the body. As for all error so I suppose we are good with the current return value

Yes, in general for anything other than success i.e 2xx status, it checks the body for the error msg.

mfvitale:
@pxcamus Could you solve the conflicts?

pxcamus:
@mfvitale all passing now

mfvitale (approved):
LGTM! Thanks!

@indraraj Is this safe to be merged without changes on the FE?

indraraj:
Yes, it's fine to merge.

mfvitale:
@pxcamus Applied, thanks!

PR #530, opened 2026-09-01
debezium/dbz#1963 Allow configuring message format via source additional properties

alinemoreirar:
## Description

Conductor currently ignores any `debezium.format.*` properties passed through
a source's additional properties: `PipelineMapper#createSource` forwards the
whole `source.getConfig()` map into `spec.source.config`, which the Debezium
Server engine emits as `debezium.source.debezium.format.*` — a namespace it
never reads. There is no other way today to set `spec.format` on the
generated `DebeziumServer` CR through the Platform API/UI, even though the
CRD (and `io.debezium.operator.api.model.Format`) already support it.

This affects at least two real use cases raised in the issue:
- Configuring an Avro converter against a schema registry (original report).
- Disabling the JSON schema envelope (`schemas.enable=false`) per pipeline
  (our case).

As agreed with @mfvitale in the issue thread, this implements the proposed
interim solution: let users set `debezium.format.(key|value|header)` and
`debezium.format.(key|value|header).*` as source additional properties, and
have Conductor recognize and lift them into `spec.format` on the generated
CR — instead of forwarding them (incorrectly) into `spec.source.config`.

### Example

Given a source with additional properties:

```json
{
  "debezium.format.key": "json",
  "debezium.format.key.schemas.enable": "false",
  "debezium.format.value": "json",
  "debezium.format.value.schemas.enable": "false"
}
```

The generated `DebeziumServer` CR now gets:

```yaml
spec:
  format:
    key:
      type: json
      config:
        schemas.enable: "false"
    value:
      type: json
      config:
        schemas.enable: "false"
```

...and none of the `debezium.format.*` keys leak into `spec.source.config`
anymore.

### Implementation notes

- `PipelineMapper#createSource` now returns a `SourceMapping(Source, Format)`
  record instead of just `Source`, so `Format` can be built from the same
  additional-properties map without a second pass or an instance field on the
  (`@ApplicationScoped`, effectively shared) mapper.
- The extraction (`extractFormat`) mutates a defensive copy of the additional
  properties map, removing every matched key so it's never also emitted
  under `spec.source.config`.
- Backward compatible: when no `debezium.format.*` property is set,
  `spec.format` keeps its current default (`type: json`, empty `config`),
  same as today.
- This is intentionally scoped as the interim fix discussed in the issue,
  not the longer-term first-class/reusable Format resource idea — happy to
  help with that follow-up separately if useful.

### Testing

- `mvn compile` — builds clean.
- `mvn test -Dtest=PipelineMapperTest` — 15/15 pass, including two new tests:
  `testMapper_ShouldExtractFormatFromSourceAdditionalProperties` and
  `testMapper_ShouldDefaultFormatWhenNotConfigured`.
- `mvn test` (full `debezium-platform-conductor` module) — 268/268 pass, no
  regressions.

## PR Checklist
- [x] I have read the contribution guidelines and the governance document on PR expectations.

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
@@ -23,6 +23,8 @@ import io.debezium.operator.api.model.ConfigProperties;
 import io.debezium.operator.api.model.DebeziumServer;
 import io.debezium.operator.api.model.DebeziumServerBuilder;
 import io.debezium.operator.api.model.DebeziumServerSpecBuilder;
+import io.debezium.operator.api.model.Format;
+import io.debezium.operator.api.model.FormatType;
 import io.debezium.operator.api.model.Predicate;
 import io.debezium.operator.api.model.PredicateBuilder;
 import io.debezium.operator.api.model.Quarkus;
@@ -58,6 +60,8 @@ public class PipelineMapper {
 
     private static final String SIGNAL_ENABLED_CHANNELS_CONFIG = "signal.enabled.channels";
     private static final String NOTIFICATION_ENABLED_CHANNELS_CONFIG = "notification.enabled.channels";
+    private static final String FORMAT_CONFIG_PREFIX = "debezium.format.";
+    private static final List<String> FORMAT_PARTS = List.of("key", "value", "header");
     private static final String DEFAULT_SIGNAL_CHANNELS = "source,in-process";
     private static final String DEFAULT_NOTIFICATION_CHANNELS = "log";
     private static final String PREDICATE_PREFIX = "p";
@@ -101,7 +105,7 @@ public class PipelineMapper {
 
         var dsRuntime = createRuntime();
 
-        var dsSource = createSource(pipeline);
+        var dsSourceMapping = createSource(pipeline);
 
         var dsSink = createSink(pipeline);
 
@@ -118,8 +122,9 @@ public class PipelineMapper {
         var specBuilder = new DebeziumServerSpecBuilder()
                 .withQuarkus(dsQuarkus)
                 .withRuntime(dsRuntime)
-                .withSource(dsSource)
+                .withSource(dsSourceMapping.source())
                 .withSink(dsSink)
+                .withFormat(dsSourceMapping.format())
                 .withTransforms(transformations)
                 .withPredicates(predicates);
 
@@ -207,7 +212,15 @@ public class PipelineMapper {
                 .build();
     }
 
-    private Source createSource(PipelineFlat pipeline) {
+    /**
+     * Result of mapping a pipeline's source: the {@link Source} spec itself, plus the
+     * {@link Format} extracted from its {@code debezium.format.*} additional properties
+     * (see {@link #extractFormat(Map)}).
+     */
+    private record SourceMapping(Source source, Format format) {
+    }
+
+    private SourceMapping createSource(PipelineFlat pipeline) {
 
         var source = pipeline.getSource();
         var sourceConfig = new ConfigProperties();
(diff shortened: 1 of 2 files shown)


debezium-platform-conductor/src/test/java/io/debezium/platform/environment/operator/PipelineMapperTest.java
```diff
@@ -235,6 +235,52 @@ public void testMapper_ShouldNotSetOtelIntervalsWhenDisabled() {
         assertThat(otel.isEnabled()).isFalse();
     }
 
+    @Test
+    @FixFor("debezium/dbz#1963")
+    public void testMapper_ShouldExtractFormatFromSourceAdditionalProperties() {
+        var pipeline = mockPipelineWithSource(ConnectionEntity.Type.POSTGRESQL, Map.of(
```
mfvitale: Could you add another `Map` as parameter of `mockPipelineWithSource` so you can pass the config directly and do the mocks internally?

mfvitale:
@alinemoreirar Thanks for the PR. The changes LGTM overall. I just left a minor comment.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
```diff
@@ -219,16 +232,62 @@ private Source createSource(PipelineFlat pipeline) {
...
+            }
+
+            additionalProperties.entrySet().removeIf(entry -> {
+                if (!entry.getKey().startsWith(nestedPrefix)) {
+                    return false;
+                }
+                String nestedKey = entry.getKey().substring(nestedPrefix.length());
+                formatType.getConfig().setProps(nestedKey, entry.getValue());
+                return true;
+            });
```
Naros: I'll defer to @mfvitale here, but a `removeIf` predicate that mutates other state seems unusual to me and is generally probably not something one would expect in this context. Personally, I think a traditional for-loop works better when you need mutation beyond what the predicate implies.

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/operator/PipelineMapperTest.java
alinemoreirar: I'll add an overload of `mockPipelineWithSource` that takes the source's additional-properties Map and stubs `source.getConfig()` internally, so it follows the same pattern as the rest of the class instead of stubbing it inline in the test.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
alinemoreirar: I'll switch to an explicit Iterator loop over `additionalProperties.entrySet()` so the mutation (`formatType.getConfig().setProps(...)`) and the removal (`it.remove()`) are both explicit in the loop body, instead of inside the predicate.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
```diff
@@ -219,16 +233,63 @@ private Source createSource(PipelineFlat pipeline) {
...
+    private static Format extractFormat(Map<String, Object> additionalProperties) {
+        var format = new Format();
+        Map<String, FormatType> formatTypesByPart = Map.of(
+                "key", format.getKey(),
+                "value", format.getValue(),
+                "header", format.getHeader());
+
+        for (String part : FORMAT_PARTS) {
+            FormatType formatType = formatTypesByPart.get(part);
+            String typeKey = FORMAT_CONFIG_PREFIX + part;
```
mfvitale, line 273: Why to not use this instead of `FORMAT_PARTS` we know what parent configurations we are looking for right?  `debezium.format.key`, `debezium.format.value` and `debezium.format.header`.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
```diff
@@ -219,16 +233,63 @@ private Source createSource(PipelineFlat pipeline) {
...
+            String nestedPrefix = typeKey + ".";
+
+            Object type = additionalProperties.remove(typeKey);
+            if (type != null) {
+                formatType.setType(type.toString());
+            }
+
+            Iterator<Map.Entry<String, Object>> iterator = additionalProperties.entrySet().iterator();
```
mfvitale, line 281: We are doing this even if the parent configuration do not exists in the `additionalProperties`

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/operator/PipelineMapper.java
```diff
@@ -219,16 +233,63 @@ private Source createSource(PipelineFlat pipeline) {
...
+                Map.Entry<String, Object> entry = iterator.next();
+                if (entry.getKey().startsWith(nestedPrefix)) {
+                    String nestedKey = entry.getKey().substring(nestedPrefix.length());
+                    formatType.getConfig().setProps(nestedKey, entry.getValue());
+                    iterator.remove();
+                }
+            }
+        }
+
+        return format;
```
mfvitale, line 292: Could we use the `FormatBuilder`?

PR #539, opened 2026-09-06
debezium/dbz#2094 Add Host Agent module and Conductor REST integration

### LINKED ISSUE #2094: [Host Deployment ~7] Implement the Host Agent -- lightweight remote-side HTTP service
Implement the Host Agent: a lightweight Quarkus-based HTTP server deployed on each remote host that translates REST calls from the Conductor into Docker commands. Defined in [DDD-41 Section 10](https://github.com/d1vyanshu-kumar/debezium-design-documents/blob/2f320972d85d331b17eff3cb9ccbbbfb2d13f98c/DDD-41.md#10-host-agent-remote-side).

### What

The Host Agent is the host-path equivalent of the Debezium Operator -- it is like the "person on the other side" that receives instructions from the Conductor and manages Docker containers on the remote host.

- **New Maven submodule** (`debezium-platform-host-agent`) inside the `debezium-platform` repository. It compiles into its own independent, small-footprint binary (no JPA, no K8s client, no Debezium Engine dependencies).
- Single-binary HTTP server listening on a configurable port (default: `8090`).
- Deployed as a `systemd` service during Ansible provisioning (auto-starts on boot, auto-restarts on crash via `Restart=always`).
- Secured with a per-host bearer token generated during provisioning -- every request requires a valid `Authorization: Bearer <token>` header, rejected with `401 Unauthorized` otherwise.
- Uses `ProcessBuilder` to execute Docker CLI commands (same approach as Ansible execution in the Conductor).
- Deploy: receives `application.properties` content, container name, image version, and allocated port. Writes properties to `/opt/debezium/configs/<containerName>/application.properties` and bind-mounts it into the Debezium Server container.
- Deploy returns `202 Accepted` immediately so the Conductor's single-threaded Watcher is not blocked while Docker pulls images or starts containers.
- Status: computes SHA-256 hash of the deployed `application.properties` file and returns it for drift detection by the Conductor.

### REST API Endpoints
The agent must expose the following endpoints:
- `POST /api/agent/deploy` -- Starts the container (Returns 202 Accepted)
- `POST /api/agent/undeploy` -- Removes the container (docker rm -f)
- `POST /api/agent/stop` -- Stops the container
- `POST /api/agent/start` -- Starts a stopped container
- `GET /api/agent/status/{containerName}` -- Returns running status and config hash
- `GET /api/agent/logs/{containerName}` -- Returns plain text Docker logs

### Verification
- **Security Test:** Verify that calling the endpoints without a valid `Authorization: Bearer <token>` header is rejected with a `401 Unauthorized` error.
- **Endpoint Tests:** Write tests for ALL exposed endpoints (`/deploy`, `/undeploy`, `/status`, `/logs`, etc.) to verify they return the expected HTTP status codes (e.g., `202 Accepted` for deploy, `200 OK` for status).
- **Mocking:** Mock the underlying `ProcessBuilder` logic in the tests to simulate Docker command execution. This ensures the tests run cleanly without requiring a real Docker daemon on the CI build server.

Tracks a sub-task of #2085 
cc @mfvitale @kmos

d1vyanshu-kumar:
Closes: debezium/dbz#2094

### Description

Implement the **Host Agent** - a lightweight Quarkus HTTP server that runs on each remote host and manages Docker containers on behalf of the Conductor.

#### What changed

**New module: `debezium-platform-host-agent`**
- Standalone Quarkus application packaged as an uber-jar
- REST API (`/api/agent/*`) with deploy, undeploy, stop, start, status, and logs endpoints
- Bearer token authentication via `AgentTokenFilter` with constant-time comparison
- `DockerCommandRunner` executes Docker CLI via `ProcessBuilder` (sync and async via virtual threads)
- Config hash computation for drift detection

**Conductor integration**
- `HostAgentApi` - MicroProfile REST Client with `@Url` for per-host dynamic routing
- `HostAgentClient` - service wrapper handling URL construction, auth headers, and error mapping
- `AgentContainerRuntime` - `@Alternative @Priority(1)` implementation of `HostContainerRuntime` that delegates to the Agent REST API instead of Ansible
- `HostDeploymentStatusPoller` rewritten to query Agent's `/status/{name}` endpoint with exponential backoff retry
- `HostDeployment` Blaze view extended with hostname, agentPort, agentToken mappings

#### Tests
- **Host Agent**: 10 endpoint tests + 5 auth filter tests (all pass)
- **Conductor**: 13 status poller tests covering all state transitions (all pass)

#### How it works
```
Conductor → HTTP POST → Agent → docker run ...
(no more SSH/Ansible for container lifecycle)
```

The `HostPipelineController` is unchanged, CDI automatically injects `AgentContainerRuntime` instead of `AnsibleContainerRuntime`.


debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -206,6 +206,27 @@ public List<HostDeployment> findByStatus(DeploymentStatus status) {
...
+     * when the {@link HostContainerRuntime} interface only provides the
+     * container name string.
+     *
+     * @return the deployment, or {@code null} if not found
+     */
+    @Transactional(SUPPORTS)
+    public HostDeployment findByContainerName(String containerName) {
+        return evm.applySetting(
+                EntityViewSetting.create(HostDeployment.class),
+                cb().where("containerName").eq(containerName))
```
mfvitale: Could you please add a constant for the column name?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
```diff
@@ -206,6 +206,27 @@ public List<HostDeployment> findByStatus(DeploymentStatus status) {
...
+                .getResultList()
+                .stream()
+                .findFirst()
+                .orElse(null);
```
mfvitale: WDYT to return an Optional?

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+ *   <li>{@code POST /stop}     — gracefully stops a container</li>
+ *   <li>{@code POST /start}    — starts a stopped container</li>
+ *   <li>{@code GET  /status/{name}} — container running state + config hash</li>
+ *   <li>{@code GET  /logs/{name}}   — last 500 lines of container logs</li>
+ * </ul>
+ *
+ * @see DockerCommandRunner
+ * @see AgentTokenFilter
+ */
+@jakarta.ws.rs.Path("/api/agent")
```
mfvitale: Could you add the relative import and remove then the full qualification?

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+     * created by the Host Agent (running as {@code root}) and immediately
+     * reassigned to {@code 185:0} (the {@code jboss} user inside the official
+     * {@code quay.io/debezium/server} image) via {@code chown}. The container
+     * is started with {@code --user 185:0} so the Debezium Server process
+     * runs as UID 185, which owns the data directory and can write
+     * {@code offsets.dat} and {@code schema-history.dat} without requiring
+     * world-writable ({@code 777}) permissions.
+     */
+    @POST
+    @jakarta.ws.rs.Path("/deploy")
```
mfvitale: Same as above.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+            Files.createDirectories(Path.of(dataDir));
+
+            // Assign data directory ownership to UID 185 (Debezium Server jboss user).
+            // The Host Agent runs as root so it can create /opt/debezium/data/, but the
+            // Debezium Server container runs as UID 185. Without chown, UID 185 gets
+            // only 'r-x' (group/other bits of a root-owned 755 dir) and cannot write
+            // offsets.dat or schema-history.dat, causing AccessDeniedException.
+            // Setting owner=185:0 keeps mode 0755 while giving UID 185 write access
+            // as the directory owner — no insecure world-write (777) needed.
+            new ProcessBuilder("chown", "185:0", dataDir)
```
mfvitale: Could you add a constant for `"185:0"`?

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+     * container does not exist.
+     */
+    @POST
+    @jakarta.ws.rs.Path("/undeploy")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response undeploy(ContainerNameRequest request) {
+        String containerName = request.containerName();
+        logger.infov("Undeploy request received for container {0}", containerName);
+
+        docker.run("docker", "rm", "-f", containerName);
```
mfvitale: Should we stop it before?

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+     * Returns the running state and config hash for a container.
+     *
+     * <p>Returns {@code 404 Not Found} if the container does not exist,
+     * allowing the Conductor to distinguish "container was removed" from
+     * "container exists but is stopped."
+     */
+    @GET
+    @jakarta.ws.rs.Path("/status/{containerName}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response status(@PathParam("containerName") String containerName) {
```
mfvitale: Seems you used different approach to pass the `container name`. Here you used the path param, on the other you used the `ContainerNameRequest` and the there is the `DeplpymentRequest`. Let's try to uniform. We can keep the `DeplpymentRequest` and for the others just use the `@PathParam("containerName")` WDYT?

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+                    .build();
+        }
+    }
+
+    /**
+     * Force-removes a container. Idempotent — succeeds even if the
+     * container does not exist.
+     */
+    @POST
+    @jakarta.ws.rs.Path("/undeploy")
```
mfvitale: Same as above.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+        cleanupContainerFiles(containerName);
+
+        return Response.ok().build();
+    }
+
+    /**
+     * Gracefully stops a running container.
+     */
+    @POST
+    @jakarta.ws.rs.Path("/stop")
```
mfvitale: Same as above.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        return Response.ok().build();
+    }
+
+    /**
+     * Starts a previously stopped container.
+     */
+    @POST
+    @jakarta.ws.rs.Path("/start")
```
mfvitale: Same as above.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * Returns the running state and config hash for a container.
+     *
+     * <p>Returns {@code 404 Not Found} if the container does not exist,
+     * allowing the Conductor to distinguish "container was removed" from
+     * "container exists but is stopped."
+     */
+    @GET
+    @jakarta.ws.rs.Path("/status/{containerName}")
```
mfvitale: Same as above.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+        String configHash = computeConfigHash(containerName);
+
+        return Response.ok(new ContainerStatus(running, configHash)).build();
+    }
+
+    /**
+     * Returns the last 500 lines of container logs as plain text.
+     */
+    @GET
+    @jakarta.ws.rs.Path("/logs/{containerName}")
```
mfvitale: Same as above.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+
+        try {
+            byte[] content = Files.readAllBytes(configPath);
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] hash = digest.digest(content);
+            return HexFormat.of().formatHex(hash);
+        }
+        catch (IOException | NoSuchAlgorithmException e) {
+            logger.warnv(e, "Failed to compute config hash for container {0}", containerName);
+            return null;
```
mfvitale: Why not `Optional`?

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,318 @@
...
+     * {@code quay.io/debezium/server} image) via {@code chown}. The container
+     * is started with {@code --user 185:0} so the Debezium Server process
+     * runs as UID 185, which owns the data directory and can write
+     * {@code offsets.dat} and {@code schema-history.dat} without requiring
+     * world-writable ({@code 777}) permissions.
+     */
+    @POST
+    @jakarta.ws.rs.Path("/deploy")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response deploy(DeployRequest request) {
```
mfvitale: `Resource` should just validate the input, eventually map to internal object and just give the response back. All the main logic should be in a dedicated service.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/ContainerStatus.java
```diff
@@ -0,0 +1,32 @@
...
+ * the endpoint returns {@code 404 Not Found} instead of this object. This
+ * distinction allows the Conductor to differentiate "container stopped"
+ * from "container was removed."
+ *
+ * @param running     {@code true} if the container is currently running
+ * @param configHash  SHA-256 hash of the deployed config file, or {@code null}
+ *                    if the config file does not exist on disk
+ */
+public record ContainerStatus(
+        boolean running,
```
mfvitale, line 30: Why not reuse `io.debezium.platform.data.model.DeploymentStatus`?

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
d1vyanshu-kumar: Agreed! I'll update findByContainerName to return `Optional<HostDeployment>`. Fix coming right up.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/DockerCommandRunner.java
```diff
@@ -0,0 +1,112 @@
...
+ * <p>This follows the same {@code ProcessBuilder} pattern used by
+ * {@code AnsibleHostProvisioner} in the Conductor module. Each method
+ * constructs a process, captures stdout + stderr, and returns a
+ * {@link CommandOutput} with the exit code and combined output.
+ *
+ * <p>Docker commands run locally on the host where the Agent is deployed,
+ * so no SSH or Ansible is involved — just a direct {@code docker} CLI call.
+ */
+@ApplicationScoped
+public class DockerCommandRunner {
```
mfvitale, line 31: Could you apply the [command pattern](https://www.baeldung.com/java-command-pattern)?

debezium-platform-host-agent/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+        <maven.compiler.testSource>${debezium.java.source}</maven.compiler.testSource>
+        <maven.compiler.testTarget>${debezium.java.specific.target}</maven.compiler.testTarget>
+        <maven.compiler.testRelease>${debezium.java.specific.target}</maven.compiler.testRelease>
+
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+
+        <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+        <version.quarkus.platform>3.34.2</version.quarkus.platform>
```
mfvitale: Please align with the conductor version

debezium-platform-host-agent/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+                <version>${version.surefire-plugin}</version>
+                <configuration>
+                    <systemPropertyVariables>
+                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                    </systemPropertyVariables>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
```
mfvitale, line 131: Could you try to add also a native profile to have a native build?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/agent/HostAgentClient.java
```diff
@@ -0,0 +1,191 @@
...
+     * @param agentToken     the bearer token for authentication
+     * @param containerName  container name to deploy
+     * @param image          Docker image to run
+     * @param port           host port to bind
+     * @param configContent  application.properties content
+     */
+    public void deploy(String hostname, int agentPort, String agentToken,
+                       String containerName, String image, int port, String configContent) {
+        String baseUrl = buildBaseUrl(hostname, agentPort);
+        String authHeader = buildAuthHeader(agentToken);
```
mfvitale: Could you check if [custom headers](https://quarkus.io/guides/rest-client/#custom-headers-support) could help to automatically add the token?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/agent/HostAgentClient.java
```diff
@@ -0,0 +1,191 @@
...
+            Response response = agentApi.logs(baseUrl, authHeader, containerName);
+            return response.readEntity(String.class);
+        }
+        catch (WebApplicationException e) {
+            return "[Log retrieval failed: " + e.getMessage() + "]";
+        }
+    }
+
+    private static String buildBaseUrl(String hostname, int agentPort) {
+        return "http://" + hostname + ":" + agentPort;
```
mfvitale: Could you use `String.format`?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * @see HostContainerRuntime
+ * @see AnsibleContainerRuntime
+ * @see HostAgentClient
+ *
+ * @author Divyanshu Kumar Nayak
+ */
+@ApplicationScoped
+@Alternative
+@jakarta.annotation.Priority(1)
```
mfvitale: Ad the import and avoid the fully qualified name

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
```diff
@@ -0,0 +1,141 @@
...
+public class AgentContainerRuntime implements HostContainerRuntime {
```
mfvitale, line 44: The choice between the agent and the ansible one must be done via a configuration

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override
+    public void undeploy(String host, String containerName) {
+        HostDeployment deployment = deploymentService.findByContainerName(containerName);
+        if (deployment == null) {
+            logger.warnv("No deployment found for container {0} on {1}, skipping Agent undeploy", containerName, host);
+            return;
+        }
+        agentClient.undeploy(deployment.getHostname(), deployment.getAgentPort(),
+                deployment.getAgentToken(), containerName);
+    }
```
mfvitale, line 79: Where is the deletion of the `HostDeploymentEntity`?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
```diff
@@ -0,0 +1,141 @@
...
+        return agentClient.logs(deployment.getHostname(), deployment.getAgentPort(),
+                deployment.getAgentToken(), containerName);
+    }
+
+    /**
+     * Fallback hostname resolution when the deployment record is not yet available.
+     * This should rarely happen — only during the brief window between
+     * {@code createDeployment()} and the actual deploy call.
+     */
+    private String resolveHostname(String sshAlias) {
```
mfvitale: Is this really needed?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AnsibleContainerRuntime.java
```diff
@@ -22,9 +22,10 @@
...
  * {@link AnsibleCommandRunner} for the actual process invocation.
  *
- * <p>When the Host Agent replaces Ansible (sub-issue 7+), a new
- * implementation will call the agent's REST API instead.
+ * <p>Superceded by {@link AgentContainerRuntime} as the default runtime
+ * implementation, but retained for fallback and reference purposes.
  */
+
 @ApplicationScoped
 public class AnsibleContainerRuntime implements HostContainerRuntime {
```
mfvitale, line 32: This should also be used based on the configuration property that let you choose between agent and ansible mode.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostDeploymentStatusPoller.java
```diff
@@ -40,60 +41,45 @@
...
-            "no such container");
-
-    /** Remote path format for the deployed config file (matches HostPipelineController). */
-    private static final String CONFIG_PATH_FORMAT = "%s/%s/application.properties";
-    private static final String HASH_COMMAND_FORMAT = "sha256sum %s | awk '{print $1}'";
-
     private final Logger logger;
     private final HostDeploymentService deploymentService;
-    private final AnsibleCommandRunner ansibleRunner;
+    private final HostAgentClient agentClient;
```
mfvitale, line 84: In this way the poller will no longer works when `AnsibleContainerRuntime`. It must work.

debezium-platform-conductor/src/main/resources/ansible/host-setup.yml
```diff
@@ -146,17 +146,30 @@
...
         mode: "0755"
       tags: [host_agent]
 
+    - name: Copy Host Agent binary JAR
+      ansible.builtin.copy:
+        src: "{{ agent_binary_path | default(playbook_dir + '/../../../../debezium-platform-host-agent/target/debezium-host-agent-runner.jar') }}"
+        dest: /opt/debezium/agent/debezium-host-agent-runner.jar
+        mode: "0755"
+      tags: [host_agent]
+
```
mfvitale: This should be just something temp and just for testing it locally. The final playbook should get it from maven.

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/HostDeploymentStatusPollerTest.java
```diff
@@ -22,47 +23,47 @@
...
- *   <li>Transient SSH failures are retried via {@code RetryingRunnable}</li>
- *   <li>Definitive "no such object" failures are trusted immediately</li>
+ *   <li>Transient HTTP failures retried via RetryingRunnable</li>
  * </ul>
  */
 class HostDeploymentStatusPollerTest {
 
     private HostDeploymentService deploymentService;
-    private AnsibleCommandRunner ansibleRunner;
+    private HostAgentClient agentClient;
```
mfvitale, line 57: Same comment as in `HostDeploymentStatusPoller`. The poller must work with the ansible and the agent approach.

mfvitale (changes requested):

mfvitale:
@d1vyanshu-kumar Thanks for the PR. It is a good start but there are some changes required.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
d1vyanshu-kumar: Added `CONTAINER_NAME_ATTRIBUTE` and used it in the lookup so the persistence attribute is defined in one place.

debezium-platform-conductor/src/main/java/io/debezium/platform/domain/HostDeploymentService.java
d1vyanshu-kumar: Updated `findByContainerName() `to return `Optional<HostDeployment>`. The callers now handle missing deployment state explicitly instead of relying on null.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
d1vyanshu-kumar: Replaced the ownership literal with `DEBEZIUM_SERVER_USER.`

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
d1vyanshu-kumar: Added a graceful stop before removal in the Conductor undeploy flow. The Agent undeploy operation also retains its local stop attempt as protection for direct Agent calls.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
d1vyanshu-kumar: Agreed and updated the API contract. Deploy keeps `DeployRequest`; undeploy, stop, start, status, and logs now use `@PathParam("containerName").` I removed the no-longer-needed `AgentContainerNameRequest`.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
d1vyanshu-kumar: Changed the internal hash calculation to return `Optional<String>`. The nullable value is only exposed at the HTTP DTO boundary, and the poller skips drift detection when no hash is available

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
d1vyanshu-kumar: Moved Docker and filesystem lifecycle logic into `AgentContainerService`. `AgentResource` now validates input, delegates to the service, and maps results to HTTP responses.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/ContainerStatus.java
d1vyanshu-kumar: I reviewed this and kept `ContainerStatus` separate. `DeploymentStatus` represents Conductor orchestration states such as `DEPLOYING`, `FAILED`, and `CONFIG_DRIFT`, while the Agent only reports local container facts such as running state and config hash. Reusing it would also make the standalone Agent depend on the Conductor module.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/DockerCommandRunner.java
d1vyanshu-kumar: Applied the command pattern through `HostCommand`, `DockerCommand`, and `ChangeOwnershipCommand`. `DockerCommandRunner` now executes typed command objects rather than receiving arbitrary process arguments from the resource layer

debezium-platform-host-agent/pom.xml
d1vyanshu-kumar: Aligned the Host Agent Quarkus platform version with the Conductor version,` 3.38.2.`

debezium-platform-host-agent/pom.xml
d1vyanshu-kumar: Added the native profile matching the Conductor module. The current provisioning flow still installs the Java uber JAR, but the Agent can now be built consistently in native mode as well.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/agent/HostAgentClient.java
d1vyanshu-kumar: Updated this to use` @ClientHeaderParam` with the token method parameter marked as `@NotBody.` The Authorization header is now applied consistently by the REST client for every Agent endpoint.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/agent/HostAgentClient.java
d1vyanshu-kumar: Updated` buildBaseUrl() `to use` String.format("http://%s:%d", hostname, agentPort)`

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
d1vyanshu-kumar: Removed the old priority-based alternative selection entirely. Runtime selection is now property based, so `@Alternative`, `@Priority`, and the fully qualified annotation are no longer needed here.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
d1vyanshu-kumar: Agreed. Runtime selection now uses `platform.host.container-runtime`. agent selects `AgentContainerRuntime`; ansible selects `AnsibleContainerRuntime`, and remains the default for compatibility.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
d1vyanshu-kumar: The remote runtime does not delete Conductor database state. `HostPipelineController.executeUndeploy() `performs remote cleanup first and then deletes the `HostDeployment` record through `HostDeploymentService`. This keeps persistence ownership in Conductor.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AgentContainerRuntime.java
d1vyanshu-kumar: Removed it. Lifecycle operations now use the hostname, Agent port, and Agent token persisted with the deployment. This avoids deriving an HTTP endpoint from an SSH alias or falling back to incomplete data.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/AnsibleContainerRuntime.java
d1vyanshu-kumar: Updated `AnsibleContainerRuntime` to use the same property-based selection. It is selected for `ansible` and remains eligible when the property is absent.

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostDeploymentStatusPoller.java
d1vyanshu-kumar: Updated the poller to support both Agent and Ansible runtime strategies. It chooses only the runtime-specific status probe differently; status transitions, grace-period handling, and drift handling remain shared.

debezium-platform-conductor/src/main/resources/ansible/host-setup.yml
d1vyanshu-kumar: Replaced the local build-path copy with `community.general.maven_artifact.` The playbook receives the optional repository URL through configuration, while Maven Central remains the default when no override is supplied. I used the local Nexus repository only for manual verification of the snapshot artifact.

debezium-platform-conductor/src/test/java/io/debezium/platform/environment/host/HostDeploymentStatusPollerTest.java
d1vyanshu-kumar: Added coverage for both Agent and Ansible poller paths, including their runtime-specific probes and common status/drift transitions. The wiring tests also verify default Ansible selection and explicit Agent selection.

d1vyanshu-kumar:
Hi @mfvitale, thanks for the detailed review. I addressed all of the feedback in the follow-up commits and replied on the individual threads with the relevant implementation details.
In summary, the Agent and Ansible runtimes are now selected through configuration; provisioning installs the Agent only when the Agent runtime is selected, and the poller supports both runtime strategies. I also simplified the Agent REST contract, moved lifecycle logic into the dedicated service layer, kept deployment persistence owned by Conductor, and updated the Maven-based Agent provisioning flow.
I ran the targeted unit and wiring tests for the Agent, runtime selection, controller, provisioner, and both poller paths. I also completed manual end-to-end verification on `test-host,` including Nexus snapshot publishing, automatic Ansible provisioning, systemd startup, token authentication, request validation, deploy, and undeploy.

image
Manual Agent API verification on `test-host`: missing token returns 401, an authenticated request for a missing container returns 404, and an invalid container name returns 400.

debezium-platform-host-agent/pom.xml
mfvitale: Just out of curiosity: did you tried it?

debezium-platform-conductor/src/main/resources/ansible/host-setup.yml
mfvitale: Wondering if we could distribute it as OCI image and/or native executable. 

@jpechane @Naros Any thoughts on this?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/config/HostConfigGroup.java
```diff
@@ -31,6 +31,10 @@
 @ConfigRoot(phase = ConfigPhase.RUN_TIME)
 public interface HostConfigGroup {
 
+    String CONTAINER_RUNTIME_PROPERTY = "platform.host.container-runtime";
```
mfvitale: I'm wondering if we could move the `platform.deployment.mode` and `platform.host.container-runtime` and remove the prefix `platform` so effectively they becomes `pipeline.deployment.mode` and `pipeline.host.container-runtime` since now is more clear that are related to the pipelines. WDYT?

debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
```diff
@@ -234,6 +234,13 @@ private void failDeployment(Long pipelineId, String reason) {
         logger.errorv("Deployment failed for pipeline {0}: {1}", pipelineId, reason);
     }
 
+    private HostContainerRuntime containerRuntime() {
```
mfvitale: In this case I think it is safe to directly go with the `get()` since there is at least a default implementation that is the ansible one.

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,164 @@
...
+        if (request.port() < 1 || request.port() > 65_535) {
+            return Optional.of(badRequest("The container port must be between 1 and 65535"));
+        }
+        if (request.configContent() == null) {
+            return Optional.of(badRequest("Container configuration is required"));
+        }
+        return Optional.empty();
+    }
+
+    private Optional<Response> validateContainerName(String containerName) {
```
mfvitale: I think you could use jakarta validation?

debezium-platform-host-agent/src/main/java/io/debezium/platform/agent/AgentResource.java
```diff
@@ -0,0 +1,164 @@
...
+
+        try {
+            return operation.execute();
+        }
+        catch (AgentOperationException e) {
+            return operationFailed(operationName, containerName, e);
+        }
+    }
+
+    private Optional<Response> validateDeployRequest(DeployRequest request) {
```
mfvitale: I think you could use jakarta validation?

debezium-platform-host-agent/pom.xml
d1vyanshu-kumar: Not yet. I added the standard Quarkus native profile to align with Conductor, but my local testing was focused on the Java JAR installation path. I’d want to build and test the native executable separately before claiming it's fully validated.

d1vyanshu-kumar:
Hi @mfvitale , I removed the `containerRuntime()` helper completely and now use `containerRuntime.get()` directly at each lifecycle call site. Since `AnsibleContainerRuntime` remains available as the default when no runtime property is set, the helper was no longer adding any behavior.
 I also addressed the other follow-up comments by moving the deployment-selection settings to `pipeline.deployment.mode` and `pipeline.host.container-runtime`, while keeping host provisioning settings under `platform.host`. The Agent API now uses Jakarta Validation for deployment requests and container-name path parameters, so invalid requests are rejected before they reach the container service. PTAL!

mfvitale:
@d1vyanshu-kumar One last thing. Could you add open api annotations on the agent api as in conductor?

d1vyanshu-kumar:
> @d1vyanshu-kumar One last thing. Could you add open api annotations on the agent api as in conductor?

Hi @mfvitale , I added OpenAPI support and annotations to the Host Agent API, following the same pattern used by Conductor resources. Each lifecycle endpoint now documents its operation, success response, and the relevant validation, authentication, missing-container, and runtime-failure responses. I also added a focused test that verifies the generated `/q/openapi` document contains the Agent operations and response metadata.PTAL

mfvitale (approved):

mfvitale:
@d1vyanshu-kumar Great work! Applied, thanks! :tada:

PR #557, opened 2026-09-10
debezium/dbz#2584 Fix double undeploy execution and orphaned host_deployment records

### LINKED ISSUE #2584: Double undeploy execution and clean up orphaned host_deployment records during asynchronous outbox pipeline deletion
### Description
In host deployment mode, pipeline deletion currently exhibits two issues with outbox-driven asynchronous undeployment:

1. **Double undeploy execution:** `HostPipelineController.undeploySync()` executes `executeUndeploy()` synchronously during the HTTP DELETE request. When the Outbox CDC event later fires, `PipelineConsumer` calls `undeploy()` a second time, triggering redundant cleanup calls.
2. **Orphaned database records & poller log spam:** When `undeploySync()` is made a no-op to align with the Outbox pattern (matching operator mode), `PipelineService.delete()` removes the `PipelineEntity` DB row before `PipelineConsumer` processes the event. By the time `executeUndeploy()` runs on the event thread, `findByPipelineId()` returns empty because the parent `PipelineEntity` is gone. This leaves the `host_deployment` row orphaned in the database, causing `HostDeploymentStatusPoller` to repeatedly poll non-existent deployments and log `WARN` messages every 30 seconds endlessly.

d1vyanshu-kumar:
Fixes debezium/dbz#2584

## Description

In host deployment mode, pipeline deletion via the Outbox pattern triggered two separate undeploy executions:

1. `HostPipelineController.undeploySync()` ran full container cleanup   (stop + remove + DB record deletion) synchronously during the HTTP DELETE
2. The Outbox CDC event later fired `undeploy()` → `executeUndeploy()` a second time, finding no deployment record and logging a confusing message

The root cause was the `@OneToOne` mapping on `HostDeploymentEntity.pipeline` without `optional = true`, which made Hibernate generate an INNER JOIN in `findByPipelineId()`. After the pipeline row was deleted, the INNER JOIN
returned nothing — even though the `host_deployment` row still existed.

**Fix:**

Set `@OneToOne(optional = true)` and `@JoinColumn(nullable = true)` on the pipeline relationship. This makes Hibernate use a LEFT JOIN, so `findByPipelineId()` can still find the `host_deployment` row after the pipeline is deleted. With this, `undeploySync()` becomes a no-op (matching operator mode) and container cleanup is handled exclusively by the outbox-driven `executeUndeploy()` - executing exactly once.

## PR Checklist

- [X] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md)
  and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [X] Minimal changes to code not directly related to your change
- [X] One feature/change per PR
- [X] Rebased on upstream `main`

### DIFF
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/HostDeploymentEntity.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/data/model/HostDeploymentEntity.java
@@ -36,8 +36,8 @@ public class HostDeploymentEntity {
     @GeneratedValue
     private Long id;
 
-    @OneToOne
-    @JoinColumn(name = "pipeline_id", unique = true, nullable = false)
+    @OneToOne(optional = true)
+    @JoinColumn(name = "pipeline_id", unique = true, nullable = true)
     private PipelineEntity pipeline;
 
     @ManyToOne
--- a/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
+++ b/debezium-platform-conductor/src/main/java/io/debezium/platform/environment/host/HostPipelineController.java
@@ -104,7 +104,10 @@ public class HostPipelineController implements PipelineController {
 
     @Override
     public void undeploySync(Long pipelineId) {
-        executeUndeploy(pipelineId);
+        // The pipeline→host_deployment FK was dropped (V3.7.0.3) and the
+        // @OneToOne is optional, so pipeline deletion does not require
+        // prior host_deployment cleanup. Container teardown is handled
+        // exclusively by the outbox-driven undeploy(pipelineId).
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


d1vyanshu-kumar:
Hi @mfvitale, whenever you get a chance, could you take a look at this PR?

mfvitale:
@d1vyanshu-kumar I think we should resolve it in another way.

We need to configure the pipeline Id join as a left join adding `optional=true` on the `@OneToOne` and also `@JoinColumn(..., nullable = true)`

d1vyanshu-kumar:
Thanks @mfvitale, you're right; that's a very clean and simpler approach. The root cause was the `@OneToOne` generating an INNER JOIN. Fixed it by adding `optional = true` on the `@OneToOne` and `nullable = true` on the `@JoinColumn`. Updated tests accordingly. PTAL

mfvitale:
@d1vyanshu-kumar Applied, thanks!

PR #560, opened 2026-09-14
debezium/dbz#1725 Improve the initial loading state management in the Platform UI

### LINKED ISSUE #1725: Improve the initial loading state managment in the Platfrom UI
- Remove the Loading text shown in the initial load window
- For the secondary screen in case of a network issue, the UI keeps showing the loading state. Update it to show the network failure after 3 attempts.

shrsu:
fixes debezium/dbz#1725

--- 

The listing page that fails its first fetch would revert back to the loading state on every poll, so with conductor unreachable the UI would cycle between the spinner and a temporary error rather than resolve itself — while polling and attempting three times to re-fetch using react-query on a backend that is already down.

Now the polling is reused as the retry mechanism:

- The `retry` mechanism built into react-query is disabled; the polls *are* the attempts
- A failed request is retried after 1 second, and when there are `POLLING.maxFailures` of such failures in a row, the polling stops and the failure is communicated
- The failure message names the resource and provides a handy **Try again** button that resets the request budget, allowing recovery without having to reload the page
- The data that is already displayed is preserved while the attempts run
- The "Loading..." caption is removed from the spinner

Uses the existing `statusMessage:apis.connectionErrorTitle` / `connectionErrorDescription` keys, that have been translated in both languages; the only added translation string is `tryAgain`.

Also fixes a `goto` translation string on the pipelines error screen that does not exist in either language — i18next used the actual string "goto" as a button text instead. Same JSX block as the above error state fix.

Before and after video: [Improve the initial loading state managment in the Platfrom UI.mp4](https://drive.google.com/file/d/148lOz-g9pjZlHSGxe2Zd30jh3th58FPt/view?usp=sharing)

---


indraraj:
Thanks for the PR, @shrsu. This looks good. Just one suggestion could you add the ApiError component to the Alert-related pages so it would make the experience more consistent.

shrsu:
Pushed the changes @indraraj !
Updates video: [Improve the initial loading state managment in the Platfrom UI 2.mp4](https://drive.google.com/file/d/1I5x7NNCr6mLe3HZOeY1xNnA37POtZ22r/view?usp=sharing)

indraraj (approved):
Looks good, thanks for the PR

